PR #31, opened 2020-11-24
[feat] Add back end boilerplate

MikeVedsted:
Adding boilerplate to lay out structure of back end


Chiranjibichapagain (approved):
Great work Mike. Thanks!

MarinaFroes (approved):
It's a really good start. There's some inconsistency related to semi colons and quotes, but we're going to adjust `eslint` and `prettier` and it will fix the problem, I guess.

PR #46, opened 2020-12-01
[FEAT] Add delete user route

Rostislav013:
Added functionality to delete user from db.

Please, especially have a look for error handling and give suggestions to improve it.


MikeVedsted (approved):
Looks alright 👍 

However, you are sort of breaking the etiquette for PRs. 
One PR should implement one feature. In this case, you should have made a separate issue and PR for the errorhandler you added, so we could more easily keep track of what we have decided to develop, and the progress of each item. 
When using that model, it is also easier to go back and rectify any changes that are potentially causing issues at a later stage.

Rostislav013:
> Looks alright 👍 
> 
> 
> 
> However, you are sort of breaking the etiquette for PRs. 
> 
> One PR should implement one feature. In this case, you should have made a separate issue and PR for the errorhandler you added, so we could more easily keep track of what we have decided to develop, and the progress of each item. 
> 
> When using that model, it is also easier to go back and rectify any changes that are potentially causing issues at a later stage.
> 
> 

Ok, next times then. 

But what about current case, should I remove error handler  or keep for this time?

MarinaFroes (approved):
Looks good. In my opinion we should try to do the right things from the beginning. So I would say, remove the error handler and create a new PR just with it (it will be fast considering it's already done).
But maybe I'm asking too much. 😅

Chiranjibichapagain (approved):
looks good.

PR #47, opened 2020-12-01
[ADMIN] Configure eslint to match contribution guidelines

MarinaFroes:
This PR includes the following changes:
- Fix `Cannot use JSX unless the '--jsx' flag is provided.` issue:
  - Update of the contribution guidelines adding instructions on how to use the version of the workspace for TS;
  - Change `jsx` property on the `tsconfig.json` for the client from ` "jsx": "react-jsx",` to ` "jsx": "react",`;
- Fix end of the line issue related to `prettier` by adding `"endOfLine":"auto"` for prettier on the client

- Add  typescript rules for `eslint` on the client and remove some conflicting rules on the backend
- Add scripts to run `eslint` and `prettier` for back-end and front-end at the same time
- Add husky rule to format both back-end and front-end pre-commit

**WARNING**: I had to add `any` types as warning for the client, not error! Otherwise, I would not be able to commit without fixing it.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,22 +1,23 @@
 {
   "parser": "@typescript-eslint/parser",
-  "plugins": [
-    "@typescript-eslint",
-    "prettier"
-  ],
+  "parserOptions": {
+    "ecmaVersion": 2020,
+    "sourceType": "module",
+    "project": "./tsconfig.json"
+  },
   "extends": [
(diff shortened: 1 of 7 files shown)


MikeVedsted (approved):
Great work 👍

Rostislav013 (approved):

PR #48, opened 2020-12-01
[FEAT] Add event creation

Chiranjibichapagain:
Create Event functionality


src/controllers/event.ts
```diff
@@ -1,14 +1,15 @@
 import async from 'async'
 import { Request, Response, NextFunction } from 'express'
 import EventService from '../services/event'
+import {Event} from '../types'
```
Rostislav013, line 4: Could u fix imports, please

src/routers/event.ts
```diff
@@ -11,7 +11,7 @@ import {
 
 const router = express.Router()
 
-router.post('/', isAuthenticated, createEvent)
+router.post('/',isAuthenticated, createEvent)
```
Rostislav013, line 14: Add space between `,` and `isAuthenticated`

src/types.ts
```diff
@@ -4,3 +4,15 @@ export type GoogleToken = {
...
+export type Event = {
+  title: string
+  category: number
+  date: Date | string
+  time: string
+  description: string
+  max_participants: number
+  address: number
+  expires_at:Date
+  image:string
```
Rostislav013, line 17: Add space between : and types

src/services/event.ts
```diff
@@ -1,14 +1,36 @@
...
+
+const createEvent = async (event: Event) => {
+  const newEvent = await (
+        await pool.query(
+          'INSERT INTO event (title, category, date, time, description, max_participants, address, expires_at, image) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9) RETURNING *',
+          [event.title, event.category, event.date, event.time, event.description, event.max_participants, event.address, event.expires_at, event.image]
+        )
+  )
+  return newEvent.rows
 }
```
Rostislav013, line 25: Is it ok to delete one of awaits?
To make it as:
```
const createEvent = async (event: Event) => {
  const newEvent = await pool.query(
          'INSERT INTO event (title, category, date, time, description, max_participants, address, expires_at, image) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9) RETURNING *',
          [event.title, event.category, event.date, event.time, event.description, event.max_participants, event.address, event.expires_at, event.image]
      
  )
  return newEvent.rows
}
```

src/routers/event.ts
MarinaFroes: Yes, and I think prettier + eslint were supposed to fix this. I'll have to check this rule.

Rostislav013 (approved):

MarinaFroes (commented):
For the event creation, I think it's ok. 
But why does it include changes for google login as well? Were theses changes made on this same PR? If yes, we should consider moving them to a separate PR.

And about prettier and eslint, it seems you had some problems with them, is it correct?  `"jsx": "react"` changed back to `"jsx": "react-jsx"` when you tried to run it and you had problems with lack of spaces after commas and parenthesis? Please, let me know if you had issues because I'll need to  fix eslint rules. Thanks.

MarinaFroes:
> For the event creation, I think it's ok.
> But why does it include changes for google login as well? Were theses changes made on this same PR? If yes, we should consider moving them to a separate PR.
> 
> And about prettier and eslint, it seems you had some problems with them, is it correct? `"jsx": "react"` changed back to `"jsx": "react-jsx"` when you tried to run it and you had problems with lack of spaces after commas and parenthesis? Please, let me know if you had issues because I'll need to fix eslint rules. Thanks.

Oh, I think now I understand. Those changes for google login are from Nanna's PR but they weren't merged yet. Is that so?

client/tsconfig.json
```diff
@@ -17,7 +17,7 @@
     "resolveJsonModule": true,
     "isolatedModules": true,
     "noEmit": true,
-    "jsx": "react",
+    "jsx": "react-jsx",
     "noFallthroughCasesInSwitch": true
```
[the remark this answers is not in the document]
Chiranjibichapagain: this is not giving me any errors. Not sure if it is affecting the formatting or not.

Chiranjibichapagain:
yes, i had some misunderstanding about what we call an issue. As create
event should be authenticated i tried to check that as well at the same
time. Nana's new changes were not merged so i copied codes just to check
and pulled it later when merged.

src/controllers/event.ts
```diff
@@ -1,14 +1,15 @@
...
 export const findAllEvents = async (
   req: Request,
   res: Response,
   next: NextFunction
 ) => {
   try {
-    console.log('something should happen when this is called. Req: ', req)
+    res.json(await EventService.findAllEvents())
   } catch (error) {
     console.log(error)
```
MikeVedsted, line 14: Please use the error handler to actually handle the error. 
Along the lines of: 

```
catch (error) {
    next(new NotFoundError('No events found', error))
  }
```

src/controllers/event.ts
```diff
@@ -32,7 +33,30 @@ export const createEvent = async (
...
+      description,
+      max_participants,
+      address,
+      expires_at,
+      image
+    }
+
+    res.json(await EventService.createEvent(event))
   } catch (error) {
     console.log(error)
```
MikeVedsted, line 61: Again, handle the error instead of console.log

src/services/event.ts
```diff
@@ -1,14 +1,44 @@
...
+      event.category,
+      event.date,
+      event.time,
+      event.description,
+      event.max_participants,
+      event.address,
+      event.expires_at,
+      event.image
+    ]
+  )
```
MikeVedsted, line 31: Please refactor to 

```
const { title, category, date, time, description, max_participants, address, expires_at, image } = event

  const newEvent = await pool.query(
    'INSERT INTO event (title, category, date, time, description, max_participants, address, expires_at, image) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9) RETURNING *',
    [ title, category, date, time, description, max_participants, address, expires_at, image ]
  )
```

src/types.ts
```diff
@@ -1,6 +1,18 @@
...
+export type Event = {
+  title : string
+  category : number
+  date : Date | string
+  time : string
+  description : string
+  max_participants : number
+  address : number
+  expires_at : Date
+  image : string
```
MikeVedsted, line 17: did you or the linting add the spaces here? If possible, don't use the spaces. If the linting is the issue, reach out to Marina.
I think `date` and `time` can both be strings but address would also be a string since it is uuid v4, right?

MikeVedsted (changes requested):
A few things but overall good work 👍

src/types.ts
Chiranjibichapagain: rost wanted me to put the space.. :D i can remove it.

Chiranjibichapagain: about address yes UUID is a string. will fix it

Rostislav013: It was like this: `image:string`.
I told about this way:  `image: string`
I didn't mean:  `image : string`
heh

MikeVedsted: to be clear, it used to be `key:type` without any spaces. Now you have `key : type`  with space on both sides of the colon. Should be `key: type` with space only _between_ the colon and the type.

MikeVedsted: > It was like this: `image:string`.
> I told about this way: `image: string`
> I didn't mean: `image : string`
> heh

two minds, one thought 😄

MarinaFroes: I also agree with `key: type`. But it's supposed to be done automatically anyway. If you're allowed to do it in another way, I think I need to fix the rule. 🤔

Chiranjibichapagain: okay  :D understood!

MikeVedsted (approved):
Great! 👍 
Thanks for the changes.

PR #50, opened 2020-12-01
[FEAT] Add error handler

Rostislav013:
+ Added error handler to backend so we can use it to handle errors. 
+ Handled error for delete user route.


MarinaFroes (approved):
I think it's ok. Thank you for moving this to a new PR, so much better! 🏅

MikeVedsted (approved):

PR #53, opened 2020-12-02
[feat] Add dictionary

MikeVedsted:
Add dictionary to align spelling and naming across contributors.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -1,0 +1,19 @@
+// cSpell settings. 
+// DO NO CHANGE VERSION NUMBER
+// Keep in mind that cSpell only catches words longer than 3 characters. 
+// Use "words" for words that should always be considered correct. 
+// Use "flagWords" for words that should always be considered incorrect.
+    
+{
+    "version": "0.1",
+    "language": "en",
+    "words": [
+        "nannadao",
+        "MikeVedsted"
+    ],
+    
+    "flagWords": [
+        "aip",
+        "hte"
+    ]
+}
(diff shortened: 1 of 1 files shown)


MarinaFroes (approved):
Good.

cspell.json
```diff
@@ -0,0 +1,19 @@
+// cSpell settings. 
+// DO NO CHANGE VERSION NUMBER
+// Keep in mind that cSpell only catches words longer than 3 characters.
```
MarinaFroes, line 3: Reminder that the spell checker will check only spelling, not grammar. 
'DO NO CHANGE' is considered correct.

MikeVedsted: 😄

Rostislav013: What about variables name or component names? For ex NavigationBar? Navbar n so on?

MarinaFroes: From Spell checker documentation:

```
How it works with camelCase
The concept is simple, split camelCase words before checking them against a list of known English words.

camelCase -> camel case
HTMLInput -> html input -- Notice that the I is associated with Input and not HTML
snake_case_words -> snake case words
camel2snake -> camel snake -- (the 2 is ignored)
Special case will ALL CAPS words
There are a few special cases to help will common spelling practices for ALL CAPS words.

Trailing s, ing, ies, es, ed are kept with the previous word.

CURLs -> curls -- trailing s
CURLedRequest -> curled request -- trailing ed
```

But in case any word of the component name is not included, we can add it to our list.

Rostislav013 (approved):

PR #55, opened 2020-12-02
[feat] Delete event

Chiranjibichapagain:
Delete event with event_id

### DIFF
--- a/src/controllers/event.ts
+++ b/src/controllers/event.ts
@@ -81,8 +81,8 @@ export const deleteEvent = async (
   next: NextFunction
 ) => {
   try {
-    console.log('something should happen when this is called. Req: ', req)
+    await EventService.deleteEvent(req, res)
   } catch (error) {
-    console.log(error)
+    next(new NotFoundError('Event not found', error))
   }
 }
--- a/src/services/event.ts
+++ b/src/services/event.ts
@@ -1,3 +1,4 @@
+import { Response, Request } from 'express'
 import { Pool } from 'pg'
 
 import { PG_USER, PG_HOST, PG_DB, PG_PW, PG_PORT } from '../util/secrets'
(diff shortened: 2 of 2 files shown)


src/controllers/event.ts
```diff
@@ -81,8 +81,8 @@ export const deleteEvent = async (
   next: NextFunction
 ) => {
   try {
-    console.log('something should happen when this is called. Req: ', req)
+    await EventService.deleteEvent(req, res)
```
MikeVedsted, line 84: you only have to pass the event id for this.

src/services/event.ts
```diff
@@ -61,8 +62,17 @@ const updateEvent = async (eventId: string, update: string) => {
   )
 }
 
-const deleteEvent = (eventId: string) => {
-  console.log('Delete event fired for id: ', eventId)
+const deleteEvent = async(req:Request, res:Response) => {
+  const eventId = req.params.eventId
```
MikeVedsted, line 66: update this to just receive the eventId

MikeVedsted (changes requested):

src/services/event.ts
Rostislav013: `req, res `are passed here, cuz can return `res.status`

src/controllers/event.ts
Rostislav013: It depends. In this case both are sent so they can be used in services.

src/services/event.ts
Chiranjibichapagain: in that case, how do i  send a response back in case of error or success?

src/controllers/event.ts
MikeVedsted: I see that we use the same way for user, so I should probably have caught this earlier.
We can do it like this to have it aligned and potentially refactor at a later point. 

⚠️ Warning: Pseudo code ahead ⚠️ 
My thinking was that we could structure it like it can be done with e.g. Mongo. 
The below is not tested and cannot be expected to work by copy/pasting.

But something like this:

Controller: 
```
export const deleteEvent = async (
  req: Request,
  res: Response,
  next: NextFunction
) => {
  const { eventId } = req.params
  try {
    res.json(await EventService.deleteEvent(eventId))
  } catch (error) {
    next(new NotFoundError('Event not found', error))
  }
}
```
Service: 
```
const deleteEvent = async (eventId: string) => {
  return await pool.query(
    'DELETE FROM event WHERE event_id=$1',
    [eventId],
    (err) => {
      if (err) {
        throw err
      }
    }
  )
}
```
For now, let's keep Chiran's code and consider a refactor later.

MikeVedsted (approved):

MarinaFroes (approved):

src/controllers/event.ts
Rostislav013: Is it normal to use await like res.json( await.....)?

Need to keep in mind to send res status. 

Also there is a situation when trying to delete id which doesn't exist. It doesn't cause any error. In terminal it just says "Delete 0". We should handle it somehow.

Issue #25, opened 2020-11-23
[FEAT] Add update functionality for user


MikeVedsted:
**💡 I would really like to solve or include**   
It should be possible to edit information not provided by google in a user document.

**👶 How would a user describe this?**   
I would like to be able to edit my preferences, wrong information stored etc.

**🏆 My dream solution would be**  
Update route for user implemented, including controller and handler.

**:2nd_place_medal: But I'd also consider it solved if**    
N/A

**💭 If you were doing it, what would you do?**   
Create route, create handler, create controller.

**♻️ Additional context**  
N/A

**🚀 I'm ready for take off**  
Before submitting, please mark if you:  
- [x] Checked that this feature doesn't already exists  
- [x] Checked that a feature request doesn't already exists  
- [x] Went through the user flow, and understand the impact  
- [x] Made sure the request shows why it is important to users but doesn't exagerates the value

PR #57, opened 2020-12-02
[FEAT] Add update functionality for user

MarinaFroes:
This PR adds update functionality for the following user properties:
- `first_name`
- `last_name`
- `profile_image`
- `profile_text`

Let me know if I need to include `base_address`, I don't know what is that.

**I do not allow the user to change emails considering we only have google login.**

I also added type for `User`. And the error handling will need to be improved in the future.
I saw Mike's comment about refactoring Chiran's PR, please let me know if I need to refactor this one as well.

close issue #25

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -11,7 +11,9 @@
     "language": "en",
     "words": [
         "nannadao",
-        "MikeVedsted"
+        "MikeVedsted",
+        "userk",
+        "signup"
     ],    
     "flagWords": [
         "aip",
--- a/src/controllers/user.ts
+++ b/src/controllers/user.ts
@@ -71,9 +71,12 @@ export const updateUser = async (
   next: NextFunction
 ) => {
   try {
-    console.log('something should happen when this is called. Req: ', req)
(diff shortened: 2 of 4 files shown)


MikeVedsted:
> Let me know if I need to include `base_address`, I don't know what is that.
> 
> **I do not allow the user to change emails considering we only have google login.**
> 
> I also added type for `User`. And the error handling will need to be improved in the future.
> I saw Mike's comment about refactoring Chiran's PR, please let me know if I need to refactor this one as well.

For the base_address this is intended to be used for the location the user wants to base their search on, if using distance as a parameter. No need to worry about this yet as we are not yet entirely sure of the form this will ultimately take. 

Users should not be able to change their email with the current set up as they would no longer be able to log in.

For now, lets aim to do refactoring as part of the sprint review, at which point we can also pay attention to aligning error handling, whether to use indices or not for queries, and whether or not we return the error wrapped or not.

MikeVedsted (approved):

Rostislav013 (approved):

PR #64, opened 2020-12-03
[feat] Edit event

Chiranjibichapagain:
update event function

### DIFF
--- a/src/controllers/event.ts
+++ b/src/controllers/event.ts
@@ -83,9 +83,11 @@ export const updateEvent = async (
   next: NextFunction
 ) => {
   try {
-    console.log('something should happen when this is called. Req: ', req)
+    const { eventId } = req.params
+    const update = req.body
+    return res.json(await EventService.updateEvent(eventId, update))
   } catch (error) {
-    console.log(error)
+    next(new NotFoundError('Event not found', error))
   }
 }
 
--- a/src/services/event.ts
+++ b/src/services/event.ts
@@ -75,24 +75,61 @@ const findEventByCategory = async (categoryId: number) => {
   }
 }
 
-const updateEvent = async (eventId: string, update: string) => {
-  console.log(
-    'Update event fired for id: ',
-    eventId,
-    'update(should be changed from string): ',
-    update
-  )
+const updateEvent = async (eventId: string, update: Partial<Event>) => {
+  try {
+    const event: Event = await (
+      await pool.query('SELECT * FROM event WHERE event_id = $1', [eventId])
+    ).rows[0]
+    if (!event) {
+      return { error: 'user not found' }
+    }
+    const {
+      title = event.title,
+      date = event.date,
+      time = event.time,
+      description = event.description,
+      max_participants = event.max_participants,
+      expires_at = event.expires_at,
+      image = event.image
+    } = update
+
+    const updatedEvent: Event[] = await (
+      await pool.query(
+        'UPDATE event SET title = $2, date = $3, time = $4, description = $5, max_participants=$6, expires_at=$7, image=$8 WHERE event_id = $1 RETURNING *',
+        [
+          eventId,
+          title,
+          date,
+          time,
+          description,
+          max_participants,
+          expires_at,
+          image
+        ]
+      )
+    ).rows
+
+    return updatedEvent
+  } catch (error) {
+    return error
+  }
 }
 
-const deleteEvent = async(req:Request, res:Response) => {
+const deleteEvent = async (req: Request, res: Response) => {
   const eventId = req.params.eventId
-  const event = await (await pool.query('SELECT * from event WHERE event_id=$1', [eventId])).rows
+  const event = await (
+    await pool.query('SELECT * from event WHERE event_id=$1', [eventId])
+  ).rows
   if (event.length === 0) {
-    return res.status(404).json({Error:'Event not found'})
+    return res.status(404).json({ Error: 'Event not found' })
   } else {
-    await pool.query('DELETE FROM event WHERE event_id=$1', [eventId], (err) => {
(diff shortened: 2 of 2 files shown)


src/controllers/event.ts
```diff
@@ -83,9 +83,11 @@ export const updateEvent = async (
   next: NextFunction
 ) => {
   try {
-    console.log('something should happen when this is called. Req: ', req)
+    const { eventId } = req.params
+    const update = req.body
```
Rostislav013, line 87: It's probably  `const { update } = req.body`

Rostislav013 (approved):

src/controllers/event.ts
Chiranjibichapagain: i don't think update is inside the req.body.  So are you sure we can access the update like that. We can access the body by const {body} = req but that not sure like what you say.

Rostislav013: hmmm, both ways might be correct.
I think it depends how it will come from client. 
Can keep it the way u wrote

MarinaFroes: For the `updateUser`, I added an object called `update` to the body and used destructuring, but it's actually unnecessary. We can add this to the list of things that we are going to refactor: remove extra object wrappers.

src/services/event.ts
```diff
@@ -75,24 +75,61 @@ const findEventByCategory = async (categoryId: number) => {
...
+  } catch (error) {
+    return error
+  }
 }
 
-const deleteEvent = async(req:Request, res:Response) => {
+const deleteEvent = async (req: Request, res: Response) => {
   const eventId = req.params.eventId
-  const event = await (await pool.query('SELECT * from event WHERE event_id=$1', [eventId])).rows
+  const event = await (
```
MarinaFroes, line 120: I know this is from the other PR, but we can also add to the list of things to refactor later: use destructuring on line 119. `const { eventId } = req.params`

src/services/event.ts
```diff
@@ -75,24 +75,61 @@ const findEventByCategory = async (categoryId: number) => {
...
-    update
-  )
+const updateEvent = async (eventId: string, update: Partial<Event>) => {
+  try {
+    const event: Event = await (
+      await pool.query('SELECT * FROM event WHERE event_id = $1', [eventId])
+    ).rows[0]
+    if (!event) {
+      return { error: 'user not found' }
+    }
```
MarinaFroes, line 85: Is this going to be caught by the `catch` of the controller?
I thought you had to throw an error in order to it be caught. 
Like this: `throw new Error()` ?

Rostislav013: If we return here new Error() it will catch error in controller?

MarinaFroes: If you throw an error here, it will be caught by the `catch` here. But the `catch` here is just returning any errors (which will be caught by the controller `catch`). At least that's how I understood it, but I might be wrong.

MarinaFroes: If you throw an error here, it will be caught by the catch here. But the catch here is just returning any errors (which will be caught by the controller catch). At least that's how I understood it, but I might be wrong.

Chiranjibichapagain: You are right, i need to throw error up there to get a custom error message.

src/services/event.ts
```diff
@@ -75,24 +75,61 @@ const findEventByCategory = async (categoryId: number) => {
...
+
+    return updatedEvent
+  } catch (error) {
+    return error
+  }
 }
 
-const deleteEvent = async(req:Request, res:Response) => {
+const deleteEvent = async (req: Request, res: Response) => {
   const eventId = req.params.eventId
```
MikeVedsted, line 119: Destructure please

src/services/event.ts
```diff
@@ -75,24 +75,61 @@ const findEventByCategory = async (categoryId: number) => {
...
-  const event = await (await pool.query('SELECT * from event WHERE event_id=$1', [eventId])).rows
+  const event = await (
+    await pool.query('SELECT * from event WHERE event_id=$1', [eventId])
+  ).rows
   if (event.length === 0) {
-    return res.status(404).json({Error:'Event not found'})
+    return res.status(404).json({ Error: 'Event not found' })
```
MikeVedsted, line 124: We should decide if we throw or return for these cases..

MikeVedsted (approved):

nannadao (approved):

src/services/event.ts
```diff
@@ -75,24 +75,61 @@ const findEventByCategory = async (categoryId: number) => {
...
+const updateEvent = async (eventId: string, update: Partial<Event>) => {
+  try {
+    const event: Event = await (
+      await pool.query('SELECT * FROM event WHERE event_id = $1', [eventId])
+    ).rows[0]
+    if (!event) {
+      throw { error: 'Event not found' }
+    }
+    const {
+      title = event.title,
```
MikeVedsted, line 87: Could use an empty line two to separate

MikeVedsted:
I think this is overall fine but as I understand it, it will be expecting all the update values, correct? 
We should either make sure that a user cannot submit empty values then or implement checks to only update if there is a value submitted, depending on how the front end implementation will be done.

MarinaFroes:
> I think this is overall fine but as I understand it, it will be expecting all the update values, correct?
> We should either make sure that a user cannot submit empty values then or implement checks to only update if there is a value submitted, depending on how the front end implementation will be done.

When we destructure the properties from the body, if the property is not in the body (it was not provided by the user), it will default to the existing data in the database for this event, so I don't think we will have problems with empty data here (and of course we can reinforce this in the frontend). But you are absolutely right, we did not add validations when we created the event (the same for users) and this can be a problem.

PR #65, opened 2020-12-03
[feat] Setup jest for server

Rostislav013:
+ Changed npm test script
+ Installed a few packages
+ Implemented one test which should be removed later

There is an issue with eslint and jest:
```
error  Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: test/controllers/user.test.ts.
The file must be included in at least one of the projects provided
```

Currently it solved by adding `"createDefaultProgram": true` into `parserOptions` in `.eslintrc`. 
`"createDefaultProgram": true` will be depreciated later, we should find another solution.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,7 +3,8 @@
   "parserOptions": {
     "ecmaVersion": 2020,
     "sourceType": "module",
-    "project": "./tsconfig.json"
+    "project": "./tsconfig.json",
+    "createDefaultProgram": true
   },
   "extends": [
     "eslint:recommended",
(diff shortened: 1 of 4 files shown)


MarinaFroes (approved):
It seems ok. I added this to issue related to the sprint review: `Improve eslint configuration for jest to remove deprecation warnings`.

nannadao (approved):

MikeVedsted (approved):
Looks alright to me, now we just have to write some actual tests :D

PR #70, opened 2020-12-03
[feat] Authentication middleware

nannadao:
This PR allows to use authentication middleware to verify valid token.


src/controllers/user.ts
```diff
@@ -59,7 +60,8 @@ export const googleLogin = async (
   next: NextFunction
 ) => {
   try {
-    await UserService.googleLogin(req, res)
+    const { id_token } = req.body
+    return res.json(await UserService.googleLogin(id_token, res))
   } catch (error) {
     console.log(error)
```
Rostislav013, line 66: Handle errors. Check folder 'helpers'.

Rostislav013 (approved):

src/controllers/user.ts
```diff
@@ -2,6 +2,7 @@ import { Request, Response, NextFunction } from 'express'
 
 import UserService from '../services/user'
 import { NotFoundError } from '../helpers/apiError'
+import { CustomRequest } from '../types'
```
MikeVedsted, line 5: Doesn't look like you are using this here?

src/helpers/generateToken.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const generateToken = (id: string) => {
+  return jwt.sign(
+    {
+      iss: 'JoinMe',
+      sub: id
+    },
+    JWT_SECRET,
+    { expiresIn: '1d' }
+  )
```
MikeVedsted, line 13: Added this for sprint review as discussed.

src/types.ts
```diff
@@ -27,3 +29,11 @@ export type User = {
...
   created_at: Date
 }
+
+export type JwtDecoded = {
+  sub: string
+}
+
+export type CustomRequest = Request & {
+  user?: User
+}
```
MikeVedsted, line 39: CustomRequest isn't really telling what it does etc. 
How about authenticationRequest or something along those lines?

src/routers/user.ts
```diff
@@ -15,7 +15,7 @@ const router = express.Router()
 
 router.get('/', findAllUsers)
 router.get('/:userId', findUserById)
-router.get('/:userEmail', findUserByEmail)
+/* router.get('/:userEmail', findUserByEmail) */
```
MikeVedsted, line 18: Not sure why this is commented out? 😄

nannadao: Because they have the same endpoint, either findUserByEmail or findUserById can work. I just keep the find by ID for now

MikeVedsted (approved):

PR #73, opened 2020-12-07
[feat] Create button components

Chiranjibichapagain:
Resuable button component created.


client/src/components/Button/Button.scss
```diff
@@ -0,0 +1,53 @@
...
+    width:128px;
+    height:35px;
+    font-family: 'Roboto', sans-serif;
+    border-style: none;
+    border-radius: 4px;
+    font-size: 18px;
+    color:#f2f2f2;
+    transition: 0.3s ease-in-out;
+}
+
```
MarinaFroes, line 12: The buttons look good, but we should follow BEM methodology for the class names.

In that case, `button` is the block, but we have modifiers for `primary`, `secondary` and `default`.
We also need modifiers to change the width of the button and borders, right? I've noticed that the buttons on the `aside`, `form` and `header` (main page) are different. 
We need to have modifiers for all those cases, as I understand. I am not sure if we need `element` in this case, though. 

And please, remember to use `&` to add the styles for the modifiers (and elements if necessary).

```css
.button {
  /* some styles */

  &--primary {
    /* some styles */
  }

  &--secondary {
     /* some styles */
   }

  /* other modifiers */
}
```

client/src/components/Button/index.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react'
+
+import { ButtonComponent } from '../../types'
+import './Button.scss'
+
+// For Type pass in "button" | "submit" | "reset"; for text pass in the text you want in buttton;
+// for style pass in "primary" | "secondary" | "default"; for handleClick pass in the function name.
```
MarinaFroes, line 7: I think we can remove the comments because we have types and they already tell us what values we can pass to those props.

client/src/App.tsx
```diff
@@ -2,6 +2,7 @@ import React from 'react'
 
 // import Routes from './Routes'
```
MarinaFroes, line 3: Please, delete commented code. We can add it again when necessary.

client/src/types.ts
```diff
@@ -19,3 +19,13 @@ export type AppState = {
...
   }
 }
+
+// Button component types
+
+type ButtonTypes = 'button' | 'submit' | 'reset'
+export type ButtonComponent = {
+  type: ButtonTypes
+  text: string
+  style: string
```
MarinaFroes, line 29: For `style`, please add only the possible values: `primary`, `secondary` and `default`.
This way it's easier to find out what should I pass to this prop and we can't add new values.

client/src/components/Button/Button.scss
```diff
@@ -0,0 +1,53 @@
+.primary {
+    background-color: #E03600;
+    width:128px;
+    height:35px;
+    font-family: 'Roboto', sans-serif;
+    border-style: none;
+    border-radius: 4px;
+    font-size: 18px;
+    color:#f2f2f2;
```
MarinaFroes, line 9: I think it's a good idea to create variables for the colours, it's easier to maintain and we don't need to remember the hex, but only the alias. We should have a file with all the colours for the project.

client/src/components/Button/Button.scss
```diff
@@ -0,0 +1,53 @@
+.primary {
+    background-color: #E03600;
+    width:128px;
+    height:35px;
```
MarinaFroes, line 4: The buttons on the page have many different widths, right? We need to consider that on the specific modifiers.

client/src/types.ts
```diff
@@ -19,3 +19,13 @@ export type AppState = {
     isAuthenticated: boolean
   }
 }
+
+// Button component types
+
+type ButtonTypes = 'button' | 'submit' | 'reset'
+export type ButtonComponent = {
+  type: ButtonTypes
```
MarinaFroes, line 27: Do you think we will use `ButtonTypes` somewhere else? Because if not, it would be better to just have `type: 'button' | 'submit' | 'reset'`. If we do this, we can see the possible values when hovering the mouse over the props in the `Button` component.

MarinaFroes (commented):
Please, consider doing a few adjustments based on my comments bellow.

client/src/components/Button/Button.scss
```diff
@@ -0,0 +1,53 @@
+.primary {
+    background-color: #E03600;
```
MikeVedsted, line 2: let's use custom variables for the colors.

client/src/components/Button/Button.scss
```diff
@@ -0,0 +1,53 @@
...
+    width:128px;
+    height:35px;
+    font-family: 'Roboto', sans-serif;
```
MikeVedsted, line 5: let's also set global font families

MikeVedsted (changes requested):
I think Marina hits the nail on the head here regarding the guidelines. 
Let's use the functionality we actually get from scss to keep the code maintainable by separating out variables and mixins where possible.

client/src/components/Button/Button.scss
Chiranjibichapagain: i created two more modifiers; large and largeActive.

client/src/components/Button/Button.scss
```diff
@@ -0,0 +1,83 @@
...
+        height:50px;
+        font-family: $roboto;
+        border-style: none;
+        border-radius: 4px;
+        font-size: 22px;
+        color:$textDark;
+        text-align: left;
+        padding:0px 10px
+    }
+}
```
MarinaFroes, line 83: The properties that all of them have in common, stay in the `.button` block, please check example bellow (you can copy and paste it).
I think maybe the most logical way would be not having the modifier `--default`, using its properties as the `button` block and then the other modifiers would change the necessary properties. But considering it's almost done, I think we can just do this adjustments and move on to the next component.

I also added the variables for font-sizes in the example. And I've changed the `--largeActive` modifier to `--large-active`. In the contributions guidelines, we decided to use one`-` to separate words in the block, element or modifier, like this : `block-name__elem-name--mod-name`. 😄 

```
.button {
  font-family: $roboto;
  border-style: none;
  border-radius: 4px;

    &--primary {
        background-color: $primary;
        width: 128px;
        height: 35px;
        font-size: $text;
        color: $textLight;
        transition: 0.3s ease-in-out;
    }

    &--primary:hover{
        background-color: white;
        border: 1px solid $primary;
        color: $primary
    }

    &--secondary {
        background-color: $secondary;
        width: 128px;
        height: 35px;
        font-size: $text;
        color: $textDark;
        transition: 0.3s ease-in-out;
    }

    &--secondary:hover{
        background-color: white;
        border: 1px solid $secondary;
        color: $secondary
    }

    &--default {
        background-color: $default;
        width: 128px;
        height: 35px;
        font-size: $text;
        color: $textLight;
        transition: 0.3s ease-in-out;
    }

    &--default:hover{
        background-color: white;
        border: 1px solid $default;
        color: $default
    }

    &--large-active {
        background-color: $primary;
        width: 327px;
        height: 50px;
        font-size: $sub-head;
        color: $textLight;
        text-align: left;
        padding: 0px 10px
    }

    &--large {
        background-color: $secondary;
        width: 327px;
        height: 50px;
        font-size: $sub-head;
        color: $textDark;
        text-align: left;
        padding: 0px 10px
    }
} 
```

client/src/types.ts
```diff
@@ -19,3 +19,11 @@ export type AppState = {
     isAuthenticated: boolean
   }
 }
+
+// Button component types
+export type ButtonComponent = {
+  type: 'button' | 'submit' | 'reset'
+  text: string
+  style: 'primary' | 'secondary' | 'default' | 'largeActive' | 'large'
+  handleClick: () => void
```
MarinaFroes, line 28: Please, change `largeActive` to `large-active`.

MarinaFroes (changes requested):
Please, consider my comments bellow. 
If you want, you can copy and paste the example I added in your `Button.scss` file. 😄

client/src/SCSS/_config.scss
```diff
@@ -16,8 +16,4 @@ $roboto:'Roboto', sans-serif;
 // font sizes
 $heading:30px;
 $sub-head:22px;
-$text:18px
-
-
```
MikeVedsted, line 21: For some reason I can't add this comment to line 1-2 but that is where it goes..
It is only necessary to have line two, where you import both fonts. Line 1 can be removed.

MikeVedsted (approved):
Not a big fan of setting sizes with for the buttons with pixels but I think we can move on for now, and refactor some of this when addressing responsiveness.

client/src/components/Button/Button.scss
```diff
@@ -0,0 +1,51 @@
...
+        border: 1px solid $primary;
+        color: $primary
+    }
+
+    &--secondary {
+        background-color: $secondary;
+        font-size: $text;
+        color: $textDark;
+    }
+
```
MarinaFroes, line 31: You forgot the `hover` for the secondary modifier.

MarinaFroes (approved):
It seems much better, thanks. 🏅 
Please, include the hover for the secondary modifier and I think we're good to go.

Rostislav013 (approved):

Issue #74, opened 2020-12-07
[FEAT] Add Google API for addresses


MikeVedsted:
**💡 I would really like to solve or include**   
Clearly and concisely describe the problem you are trying to solve.   

I would like to add google places api for addresses to make sure we have correct addresses and can easily extract geo coordinates when we get to that. 

**👶 How would a user describe this?**   
Describe how users are affected by statements users might make or a user story.  
I would like to know the exact address for the event 

**🏆 My dream solution would be**  
Describe the best possible scenario of this being implemented.  

Google auto-conolete that looks like the forms they are used in. 
Potentially select from map.

**:2nd_place_medal: But I'd also consider it solved if**    
Describe an implementation that would fulfill the basic wish but might need more work in the future.  

Data validation for 

**🚀 I'm ready for take off**  
Before submitting, please mark if you:  
- [x] Checked that this feature doesn't already exists  
- [x] Checked that a feature request doesn't already exists  
- [x ] Went through the user flow, and understand the impact  
- [x] Made sure the request shows why it is important to users but doesn't exaggerate the value

PR #78, opened 2020-12-08
[FEAT] Add google autocomplete

MikeVedsted:
Related issue: #74 

Done:
- Set up API and billing for Google Maps JS API and Places API.
- Installed react-google-autocomplete to handle autocomplete requests
- Set up local state to display extracted values as readOnly fields

Potential improvements: 
- receive handle function as prop from parent
- styling

Current bugs/limitations:
- Users are able to enter addresses without street number. This is by design but I can look into it if desired.
- If input is given but no address is selected, hitting enter will result in an error crashing the client

Nice to know:
- Currently restricted to only searching Finland and exact addresses. Config can be expanded as needed.


MarinaFroes (approved):
It seems good. 😄 

PS. Maybe we should create an issue to find a solution for this:
> If input is given but no address is selected, hitting enter will result in an error crashing the client

Rostislav013 (approved):

Chiranjibichapagain (approved):

PR #80, opened 2020-12-08
[refactor] Move db connection out services

Rostislav013:
So I have moved db connection to src. 
There are some warns in that db connection file with any types. Will have a look at it. But doesn't affect merging.  
Db connection is taken from `https://node-postgres.com/guides/project-structure`.


src/services/user.ts
```diff
@@ -91,15 +79,13 @@ const updateUser = async (userId: string, update: Partial<User>) => {
...
   const user: User = DBResponse.rows[0]
 
   if (!user) {
     throw new Error()
   } else {
-    pool.query('DELETE FROM userk WHERE user_id = $1;', [userId], (err) => {
-      if (err) throw err
-    })
+    await db.query('DELETE FROM userk WHERE user_id = $1;', [userId])
     return { message: 'User deleted' }
```
MikeVedsted, line 89: how do we handle potential errors in cases like this? Could we maybe add `...rest` to the props of your query function?

src/db/index.ts
```diff
@@ -0,0 +1,47 @@
...
+})
+
+export default {
+  async query(text: string, params?: any[]) {
+    const start = Date.now()
+    const res = await pool.query(text, params)
+    const duration = Date.now() - start
+    console.log('executed query', { text, duration, rows: res.rowCount })
+    return res
+  },
```
MikeVedsted, line 22: will it work to add `...rest` here for when we need to add functions?

MikeVedsted (approved):
nice work

Chiranjibichapagain (approved):

src/services/user.ts
Rostislav013: Cuz we use async await here, it is not needed. Controller catches it

PR #82, opened 2020-12-08
[feat] Set db helper for jest

Rostislav013:
New tests will not work until [PR80](https://github.com/MikeVedsted/JoinMe/pull/80) is merged into this.

So dbhelper is created. 

The idea is that temporary tables created with the same names as the tables that we want to test which are in our db.

Tested with local db. NOT TESTED WITH REMOTE DB. 
For that reason, please check, that u r connected to local db and you have a copy of remote db.

To get a copy of remote db:
Run script: `heroku pg:pull postgresql-crystalline-87899 mylocaldb --app joinme-integrify` . Where `mylocaldb` can be changed .
Or make it manually. Rows are not important.

More about temporary tables `https://www.postgresqltutorial.com/postgresql-temporary-table/`.


MikeVedsted:
> To get a copy of remote db:
> Run script: `heroku pg:pull postgresql-crystalline-87899 mylocaldb --app joinme-integrify` . Where `mylocaldb` can be changed .
> Or make it manually. Rows are not important.

It is important to know that the local db cannot exists in advance for this to work. So if you want to use a name that already exists, you will have to drop that db first.

MikeVedsted (approved):
Assuming we merge the related PR as is, this looks fine. 
Make sure to apply changes from #80 here as well.

Chiranjibichapagain (approved):

Issue #63, opened 2020-12-03
[FEAT] Add create new event component


nannadao:
**💡 I would really like to solve or include**   
User should be able to create a new event

**👶 How would a user describe this?**   
I would like to create an event 

**🏆 My dream solution would be**  
User should be able to create an event according to wireframe

**:2nd_place_medal: But I'd also consider it solved if**    
N/A

**💭 If you were doing it, what would you do?**   
N/A

**♻️ Additional context**  
N/A

**🚀 I'm ready for take off**  
Before submitting, please mark if you:  
- [x] Checked that this feature doesn't already exists  
- [x] Checked that a feature request doesn't already exists  
- [x] Went through the user flow, and understand the impact  
- [x] Made sure the request shows why it is important to users but doesn't exaggerate the value

PR #93, opened 2020-12-11
[FEAT] Add front end event creation form

MikeVedsted:
Connecting front and back end for event creation

#63


client/src/types.ts
```diff
@@ -26,12 +26,45 @@ export type ButtonComponent = {
...
-  short_name: string
   types: string[]
+}
+
+export type Address = {
+  number: string
+  street: string
+  city: string
+  postalCode: string
+  country: string
```
Rostislav013, line 49: Are they all string or some are `number`?

Rostislav013 (approved):

Rostislav013:
Will it be as a modal or separate page?

client/src/types.ts
MikeVedsted: We receive them as a string in the response, yes. And to keep it "simple", we just parse the values on the back end, since we don't use them in the front.

MikeVedsted:
> Will it be as a modal or separate page?

My understanding was a separate page but it doesn't really matter in terms of this component. It will just either be put in a modal component or in a page.

nannadao (approved):

MarinaFroes (approved):

PR #103, opened 2020-12-14
Feature add navigation bar

Chiranjibichapagain:


client/src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { useState } from 'react'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+
+import { faBell } from '@fortawesome/free-solid-svg-icons'
+import { faComment } from '@fortawesome/free-solid-svg-icons'
+import { faUserAlt } from '@fortawesome/free-solid-svg-icons'
+import './Navbar.scss'
+import logo from './logo2.png'
```
Rostislav013, line 8: Move those imports upper
```
import { faBell } from '@fortawesome/free-solid-svg-icons'
import { faComment } from '@fortawesome/free-solid-svg-icons'
import { faUserAlt } from '@fortawesome/free-solid-svg-icons'
```

client/src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+        alt="logo"
+      />
+      {logged ? (
+        <div className="header__icons">
+          <FontAwesomeIcon className="header__icon" icon={faBell} />
+          <FontAwesomeIcon className="header__icon" icon={faComment} />
+          <img
+            className="header__image header__image--profile"
+            src="https://images.pexels.com/photos/220453/pexels-photo-220453.jpeg?auto=compress&cs=tinysrgb&dpr=1&w=500"
+            alt="profile"
```
Rostislav013, line 38: ehh who is this guy? heh

Rostislav013, line 38: Could u make img element please clickable? 
Onclick it should lead to a profile page ` /:userId`. Can use history hook to push.

client/src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { faBell } from '@fortawesome/free-solid-svg-icons'
+import { faComment } from '@fortawesome/free-solid-svg-icons'
+import { faUserAlt } from '@fortawesome/free-solid-svg-icons'
+import './Navbar.scss'
+import logo from './logo2.png'
+
+const Navbar = () => {
+  const [nav, setNav] = useState(true)
+  const [logged, setLogged] = useState(true)
+
```
Rostislav013, line 13: Instead of it also can be userId. Take that from cookies.
That userId we can later to pass to history hook when click on user image pic.

client/src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const changeBG = () => {
+    if (window.scrollY >= 80) {
```
Rostislav013, line 15: BG is background?

client/src/components/Navbar/index.tsx
Chiranjibichapagain: There is a modal that should pop up while clicking this image and if not image login text. so maybe we can create issues for creating login modal and profile modal. And when they are ready we can link.

client/src/components/Navbar/index.tsx
Chiranjibichapagain: yes, while scrolling it will add a background to the navbar.

client/src/components/Navbar/index.tsx
MikeVedsted: I've made an issue to have these imports centralized. I think now is the right time to do it, if we are starting to add a ton of icons

client/src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,54 @@
+import React, { useState } from 'react'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+
+import { faBell } from '@fortawesome/free-solid-svg-icons'
+import { faComment } from '@fortawesome/free-solid-svg-icons'
+import { faUserAlt } from '@fortawesome/free-solid-svg-icons'
+import './Navbar.scss'
```
MikeVedsted, line 7: Please move stylesheet to bottom of imports

client/src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+import logo from './logo2.png'
+
+const Navbar = () => {
+  const [nav, setNav] = useState(true)
```
MikeVedsted, line 11: This is not very descriptive. How about `showNav` instead?

client/src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+  const [logged, setLogged] = useState(true)
```
MikeVedsted, line 12: Doesn't look like we need to use `useState` here, since we so not update it. 
If so, please refactor to a constant.

client/src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  const changeBG = () => {
```
MikeVedsted, line 14: This doesn't actually change the background, does it? 
It looks like it toggles the active modifier for the nav, so name accordingly.

client/src/components/Navbar/index.tsx
MikeVedsted: This can be written short ;)

client/src/components/Navbar/index.tsx
MikeVedsted: I've created amodal and made a pr. Once merged, it will take a component to show in a modal

client/src/components/Navbar/index.tsx
MikeVedsted, line 8: I think we should just have an svg of the logo, so it will always have the right size instead of the different versions. 
If you dont know how to do that, just send me the original logo and I can make one.

client/src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+    if (window.scrollY >= 80) {
+      setNav(false)
+    } else {
+      setNav(true)
+    }
+  }
```
Rostislav013, line 20: The shorter way `(window.scrollY >= 80) ? setNav(false) : setNav(true)`

client/src/components/Navbar/index.tsx
MikeVedsted: Oh, and it shouldn't be in this directory but in the `Assets` I suppose

client/src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  window.addEventListener('scroll', changeBG)
+
+  return (
+    <header className={nav ? 'header' : 'header header--active'}>
```
nannadao, line 25: Should it be `nav` tag instead of `header`?

client/src/components/Navbar/Navbar.scss
```diff
@@ -0,0 +1,82 @@
...
+  &__image {
+    cursor: pointer;
+
+    &--logo {
+      height: 60px;
+    }
+
+    &--profile {
+      width: 50px;
+      height: 50px;
```
MikeVedsted, line 27: this makes the profile image stretched, since you are just setting it as source. 
if you add `object-fit: cover;` it will be cropped and won't stretch

client/src/components/Navbar/index.tsx
Chiranjibichapagain: when we scroll nav background would change and when a person is logged in logged state is change and we show icons otherwise just Login text. so i thought i we need states here. Or i am not getting it..

client/src/components/Navbar/index.tsx
Chiranjibichapagain: i think we can also use header, they also use header and footer stuff. i think they use a navbar inside the header. Correct me if wrong.

MarinaFroes: I personally would use `nav` here and the `header` for the hero image and other stuff, based on the definitions provided on MDN documentation.
>  \<nav> element represents a section of a page whose purpose is to provide navigation links

> \<header> element represents introductory content, typically a group of introductory or navigational aids

But I guess we can also consider this links as navigational aids. In that case, what is the element for the hero image then?

client/src/components/Navbar/index.tsx
MikeVedsted: I mean that `logged` is never changed in this component. So there is no need to use `useState` as there is nothing updating the state. 
The component will re-render before `logged` is ever changed, so it can be a constant here but we can also just change this when injecting redux

client/src/components/Navbar/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { faBell, faComment } from '@fortawesome/free-solid-svg-icons'
+
+import logo from '../../Assets/logo.svg'
+import './Navbar.scss'
+
+const Navbar = () => {
+  const [navBg, setNavBg] = useState(true)
+  const [cookies, setCookie] = useCookies(['x-auth-token'])
+  const decodedToken = jwt.decode(cookies['x-auth-token'])
+  const userId = decodedToken?.sub
```
Rostislav013, line 15: We did this with Chiran, so can get userId from cookies n pass it in Link.
Feel free to offer other solutions to get that user id

Rostislav013 (approved):

nannadao (approved):

Issue #102, opened 2020-12-14
[FEAT] Add modal component


MikeVedsted:
**💡 I would really like to solve or include**   
Clearly and concisely describe the problem you are trying to solve.   
To reduce redirects, page reloads and general amount of pages, I'd like to add a reusable modal component to show forms and information.

**👶 How would a user describe this?**   
Describe how users are affected by statements users might make or a user story.  
"This seems old school. Where are the cool popups?"
"Why is it taking so long to switch between pages? 😢"
  
**🏆 My dream solution would be**  
Describe the best possible scenario of this being implemented.  
A reusable modal component, that pops up centered on the page, and fits the content, which ideally could be forms, information and really any other kind of content.

**🚀 I'm ready for take off**  
Before submitting, please mark if you:  
- [x] Checked that this feature doesn't already exists  
- [x] Checked that a feature request doesn't already exists  
- [x] Went through the user flow, and understand the impact  
- [x] Made sure the request shows why it is important to users but doesn't exaggerate the value

PR #105, opened 2020-12-14
[FEAT] Add modal

MikeVedsted:
Added component to open content and other components in a modal. 

For the eventform to work in a modal we will have to adjust the async functions as they are keeping the component from mounting correctly (... as I currently understand it.)

Related Issue: #102 
modal


client/src/App.tsx
```diff
@@ -1,16 +1,26 @@
...
-      <GoogleUserLogin />
+      {closeModal && (
+        <Modal
+          closeModal={() => setCloseModal(false)}
+          component={<GoogleUserLogin />}
+        />
+      )}
+      <button onClick={() => setCloseModal(true)}>
+        Click here to try out an amazing modal!
+      </button>
```
Rostislav013, line 23: Does it look like we need to change here naming to `openModal` or change `true / false` in `closeModal`? 
In line 21 I thought it closes the modal.

Rostislav013 (approved):

cspell.json
```diff
@@ -2,21 +2,22 @@
...
         "signup",
         "userk",
-        "Waterpolo"
-        "cookieparser",
-        "lusca",
-        "middlewares",
-        "resave",
-        "SAMEORIGIN",
+        "Waterpolo",
         "xframe"
```
MarinaFroes, line 21: Did you mean to remove those words?

package.json
```diff
@@ -8,6 +8,8 @@
     "client": "cd client && npm run start",
     "server": "nodemon ./src/server.ts",
     "dev": "concurrently --kill-others-on-fail \"npm run server\" \"npm run client\"",
+    "install-client": "cd client && npm install",
+    "install-all": "concurrently --kill-others-on-fail \"npm install\" \"npm run install-client\"",
```
MarinaFroes, line 12: Good idea. I was doing this twice all the time.

MarinaFroes (approved):
I haven't seen it, but the code seems ok.

client/src/components/Modal/Modal.scss
```diff
@@ -0,0 +1,35 @@
...
+    width: 60%;
+    border-radius: 8px;
+    padding: 40px;
+    background-color: white;
+    position: relative;
+  }
+
+  &__close-button {
+    background-color: #00000000;
+    color: $primary;
```
MarinaFroes, line 26: Hex with 8 numbers? Is this supposed to be black?
Can you also use `black` or  `#000`?

cspell.json
MikeVedsted: I think this is just because I had them in a different order locally. The words that were removed are all added as well.

client/src/components/Modal/Modal.scss
MikeVedsted: The additional to characters are for the alpha channel (100% transparent). 
I will change this to `#ffffff00` though, so it will fall back to white if alpha isn't support and we can maintain contrast.

cspell.json
MikeVedsted: I think this is just because I had them in a different order locally. The words that were removed are all added as well.

MikeVedsted:
> I haven't seen it, but the code seems ok.

I've added a screenshot for you now :)

MarinaFroes:
> I've added a screenshot for you now :)

Cool, thank you. 😄  Still approved.

PR #107, opened 2020-12-14
[FEAT] Add display event component

Chiranjibichapagain:


client/src/components/Event/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+  image,
+  title,
+  date,
+  time,
+  address,
+  participants,
+  max_participants,
+  description
+}: EventType) => {
+  const [details, setDetails] = useState(true)
```
MikeVedsted, line 27: Inital should be false, right? We want the dialog to start out closed.

client/src/components/Event/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+  const changeView = () => {
+    if (details) {
+      setDetails(false)
+    } else {
+      setDetails(true)
+    }
+  }
```
MikeVedsted, line 35: Since it only toggling, this is enough:
```
  const changeView = () => {
    setDetails(!details)
  }
```

client/src/components/Event/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+      return years + 'y'
+    }
+  }
+  const eventCreatedBefore = calculateTime()
+
+  return (
+    <div className="event">
+      <title className="title">
+        <p className="title__text title__text--head">{title}</p>
+        <p className="title__text title__text--time">{eventCreatedBefore}</p>
```
MikeVedsted, line 59: You can just call `calculateTime ` in curly brackets and get rid of the the extra constant

MikeVedsted (commented):
Your BEM methodoly is kind of off, and should use kebab-case, instead of camelCase.
I think it would be better and easier for you to structure your BEM if you separate this into smaller components, e.g.: 
- title
- image
- details (icons and associated text)
- buttons
- ... at some point the comment functionality

client/src/components/Event/Event.scss
```diff
@@ -0,0 +1,131 @@
...
+      color: $textDim;
+      margin-left: 10px;
+    }
+  }
+}
+
+@media only screen and (max-width: 600px) {
+  .event {
+    &__dataBox {
+      flex-direction: column;
```
MarinaFroes, line 109: Maybe we'll have to adjust the breaking points. I thought the buttons are too large at certain point. But we can keep it like this now and refactor when we have the complete page.

client/src/components/Event/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+      const days = (hours / 24).toFixed(0)
+      return days + 'd'
+    } else if (hours > 730 && hours <= 8640) {
+      const months = (hours / 730).toFixed()
+      return months + 'm'
+    } else {
+      const years = (hours / 8640).toFixed(0)
+      return years + 'y'
+    }
+  }
```
MarinaFroes, line 52: This `calculateTime` can be a helper function and you just import it here. 

Maybe renaming would be good as well, it took me sometime to understand what time this function was calculating.

Btw, this function didn't work for me because it calculates the time based on the `created_at` props and I just passed it a string that is not valid for this case. So maybe we need to improve type checking for event data (right now, I can set any string to `created_at`).

client/src/components/Event/Event.scss
```diff
@@ -0,0 +1,131 @@
...
+  &__infoText {
+    font-size: $text;
+    margin: 0px;
+    font-family: $nunito;
+    color: $textDark;
+
+    &--icon {
+      color: $secondary;
+      margin-right: 15px;
+    }
```
MarinaFroes, line 39: If we could centralize the icons that would be great. The participants icon is not centralized with the rest because it is slightly wider.

client/src/components/Event/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { faUsers } from '@fortawesome/free-solid-svg-icons'
+
+import Button from '../Button'
+import FormInputFiled from '../FormInputField'
+import { EventType } from '../../types'
+import './Event.scss'
+
+const Event = ({
+  event_id,
+  created_by,
```
MarinaFroes, line 16: It seems we're not using event_id anywhere here.

MarinaFroes (commented):
Maybe we can agree that when we add a new component, we also import it into `App.tsx` passing some hard code via props. It makes it easier for the reviewer to pull the code and check the result. Right now I'm doing that manually to be able to see the component and it takes a lot of time.
If you do that, you can remove it from `App` after the approval, right before merging it.

Or, even simpler, **add a screenshot of the component in the PR**. This way we can have an idea of what the component look like. 😄

client/src/components/Event/index.tsx
MikeVedsted: Maybe `calculateTimeSinceCreated` (a bit long) or `calculateEventAge` or something like that?

MarinaFroes: Yes, good examples. Or even using `get` instead of `calculate`. `getTimeSinceCreated` or `getEventAge`.
Any of these options are fine by me.

client/src/components/Event/index.tsx
Rostislav013: Quotes should be single '...'. 
After merging eslint PR it should complain about it.

client/src/components/Event/index.tsx
Chiranjibichapagain: yes the date needs to be in a certain format maybe. i tested it with what we get from the db and it is working

client/src/components/Event/index.tsx
Chiranjibichapagain: > Your BEM methodoly is kind of off, and should use kebab-case, instead of camelCase.
> I think it would be better and easier for you to structure your BEM if you separate this into smaller components, e.g.:
> 
> * title
> * image
> * details (icons and associated text)
> * buttons
> * ... at some point the comment functionality

image and details (icons and associated text) should come under a div so that they are flex-row. other items, iitle, buttons, comments are flex-column. Tried to think of a better way but this is what i could do.

client/src/types.ts
```diff
@@ -69,6 +69,21 @@ export type DropdownProps = {
...
+  created_at: string
+  image: string
+  title: string
+  date: string
+  time: string
+  address: string
+  participants: number
+  max_participants: number
+  description: string
+  handleAddRequest: () => void | undefined
```
MikeVedsted, line 84: I don't currently see a situation where we should accept `handleAddRequest` being undefined. Could you enlighten me?

MikeVedsted (approved):

MarinaFroes (approved):
Great. And thanks for adding the working example. 🏅

Chiranjibichapagain:
I don't remember :D maybe the typescript was throwing some error or
something. i will check.

Rostislav013 (approved):

PR #108, opened 2020-12-14
[FIX] Eslint issue on client

### LINKED ISSUE #100: [FIX] Eslint issue on client
Clearly and concisely describe the problem you are trying to solve.   

Eslint does not work properly on client. Because of that it does not let push proper commits.

**🚀 I'm ready for take off**  
- [ ] Checked that this feature doesn't already exists  
- [ ] Checked that a feature request doesn't already exists  
- [ ] Went through the user flow, and understand the impact  
- [ ] Made sure the request shows why it is important to users but doesn't exaggerate the value

MarinaFroes:
Fixes #100 

_Please, if you could check this configuration on your machine before we merge, it would be nice. I've changed a lot of things and I'm still not 100% sure it's working as intended._

### Changes proposed in this pull request:
- Changed the name of the file from `.eslintrc` to `.eslintrc.json` because the previous one is deprecated;
- Added `.scss` files to prettier check and they are also formatted on saving now;
- Added single quote rule for JSX;
- Removed `createDefaultProgram: true` from the eslint config;
- Added prettier error to client and now VS code is supposed to underline prettier errors and format on saving as it was doing for the backend
  - For this to work, it's also necessary to add the code below to your workspace configuration (file `settings.json` inside the folder `.vscode`):
```
 "eslint.workingDirectories": [
    "client"
  ]
```
- Finally, this PR also includes all the current files formatted with this new configuration.

### When reviewing this PR, please consider:
- Running this code on your machine before approving it.


Rostislav013 (approved):

MikeVedsted (approved):
I am now able to autoformat in both server and client. 
However, for server I am only getting syntax warnings, whereas for client I am also getting style warnings. That is fine to me though. 
I have not tested this for jest but the PR looks good to me 👍

MarinaFroes:
> However, for server I am only getting syntax warnings, whereas for client I am also getting style warnings. That is fine to me though.

Can you give an example of something that the client gives you warning for and the server doesn't? I would like to test it here.

MikeVedsted:
> > However, for server I am only getting syntax warnings, whereas for client I am also getting style warnings. That is fine to me though.
> 
> @MikeVedsted Can you give an example of something that the client gives you warning for and the server doesn't? I would like to test it here.

Strangest thing... I had deleted the branch after reviewing, and when I pulled this PR again, I got style warnings for both `server` and `client` 😄 
So I guess it does actually work, and I did something wrong before. 
Definitely approved by me then 👍

Issue #84, opened 2020-12-09
[FEAT] Add user profile page


Rostislav013:
**💡 I would really like to solve or include**   
Clearly and concisely describe the problem you are trying to solve.   

Add profile page for user. [Figma](https://www.figma.com/file/MAtneSrLvEXJMx27c15dn8/joinMe?node-id=0%3A1)

**👶 How would a user describe this?**   
Describe how users are affected by statements users might make or a user story.  
  
  
*Ex. Why do I have  to see the same results every time I search? Why can't it remember that I don't want to play tennis?*   

**🏆 My dream solution would be**  
Describe the best possible scenario of this being implemented.  

*Ex. User is never presented with the same activity twice.*  

**:2nd_place_medal: But I'd also consider it solved if**    
Describe an implementation that would fulfill the basic wish but might need more work in the future.  

*Ex. User can flag 'not interested' on an activity and it won't be presented to this user again.*  

**💭 If you were doing it, what would you do?**   
If possible, describe the steps you would take to implement this.   

*Ex. I would add a functionality that remembers the events that have been viewed by an individual user*   
*1. add an array to the user profile of events they have views*   
*2. check against the array before displaying an event*    

**♻️ Additional context**  
Describe any other reasons that might make this feature necessary.  

*Ex. BigCompetitor has this function and users are switching to them for this reason.*    
*Ex. We have to implement this to comply with SomeLegalStandard.*    
*Ex. MarketingTeam is unable to perform ABX until ABZ has been enabled, costing them 1 billion nanodullar per day 💰*

**🚀 I'm ready for take off**  
Before submitting, please mark if you:  
- [ ] Checked that this feature doesn't already exists  
- [ ] Checked that a feature request doesn't already exists  
- [ ] Went through the user flow, and understand the impact  
- [ ] Made sure the request shows why it is important to users but doesn't exaggerate the value

PR #112, opened 2020-12-15
[feat] User profile page

nannadao:
Issue #84


client/src/pages/ProfilePage/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+import Button from '../../components/Button'
+import Navbar from '../../components/Navbar'
+import { ProfilePageParam } from '../../types'
+import './ProfilePage.scss'
+
+const ProfilePage = ({ match }: RouteComponentProps<ProfilePageParam>) => {
+  const history = useHistory()
+  const userId = match.params.userId
+  const [profileImage, setProfileImage] = useState('')
```
Rostislav013, line 14: Is it possible to make shorter?
Like `const { userId } = match.params` ?

Rostislav013 (approved):

client/src/pages/ProfilePage/index.tsx
```diff
@@ -0,0 +1,89 @@
+import React, { useEffect, useState } from 'react'
+import axios from 'axios'
+import { RouteComponentProps } from 'react-router'
+import { useHistory } from 'react-router-dom'
```
MikeVedsted, line 4: I would have preferred to use the `useParams` hook from react-router-dom

client/src/pages/ProfilePage/index.tsx
MikeVedsted: Yes. But if it would be refactored to use `useParams` from router-router-dom, it would be `const { userId } = useParams()`

client/src/pages/ProfilePage/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+          </div>
+
+          <div className='profile__details'>
+            <h4> Email: </h4>
+            <p> {email} </p>
+          </div>
+
+          <div className='profile__details'>
+            <h4> Interests: </h4>
+            <p> {interest} </p>
```
MikeVedsted, line 71: This is going to look a bit off, when you end up receiving an array. You should iterate of the array to ensure the values display correctly.

client/src/pages/ProfilePage/ProfilePage.scss
```diff
@@ -0,0 +1,56 @@
...
+    @include flexCenter;
+    flex-direction: column;
+  }
+
+  &__left img {
+    width: 327px;
+    height: 384px;
+    margin-bottom: 26px;
+    border-radius: 8px;
+  }
```
MikeVedsted, line 19: Adding element selectors is not correct BEM methodology. 
Technically, your `left `and `right` would also be boxed, not elements.

client/src/pages/ProfilePage/ProfilePage.scss
```diff
@@ -0,0 +1,56 @@
...
+
+  &__left {
+    @include flexCenter;
+    flex-direction: column;
+  }
+
+  &__left img {
+    width: 327px;
+    height: 384px;
+    margin-bottom: 26px;
```
MikeVedsted, line 17: Would it make sense to make the column a percentage of the vw and the image relative to the column, instead of hardcoding image sizes?

client/src/pages/ProfilePage/ProfilePage.scss
```diff
@@ -0,0 +1,56 @@
...
+
+  &__header h3 {
+    font-family: $montserrat;
+    font-size: $sub-head;
+    font-weight: 600;
+    flex-basis: 25%;
+  }
+
+  &__header hr {
+    border: 1px solid #606060;
```
MikeVedsted, line 41: please change the color to a theme color

MikeVedsted (changes requested):
I don't think your routing and params are working as intended. When I try to run your code with a valid user id, I'm getting undefined responses. 
If I exchange the dynamic user id in `getUserInfo` I get the expected response and the page populates as intended. 
This can be solved by using `useParams `instead of `RouteComponentProps` and removing the `match `prop.

client/src/pages/ProfilePage/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const ProfilePage = ({ match }: RouteComponentProps<ProfilePageParam>) => {
+  const history = useHistory()
+  const userId = match.params.userId
+  const [profileImage, setProfileImage] = useState('')
+  const [fullName, setFullName] = useState('')
+  const [email, setEmail] = useState('')
+  const [interest] = useState('Sport')
+  const [profileText, setProfileText] = useState('')
+  const [address, setAddress] = useState('')
```
MikeVedsted, line 19: Since we do not expect the values to change, it is not necessary to use `useState`.

MikeVedsted (changes requested):

Rostislav013:
> I don't think your routing and params are working as intended. When I try to run your code with a valid user id, I'm getting undefined responses.
> If I exchange the dynamic user id in `getUserInfo` I get the expected response and the page populates as intended.
> This can be solved by using `useParams `instead of `RouteComponentProps` and removing the `match `prop.

After I deleted a row  in db and tried to login/create user it started to work correctly again both ways with `useParam` n match. But I prefer `useParam` too

client/src/pages/ProfilePage/index.tsx
nannadao: Please suggest an alternative. We do not have store so I have to create state to store the data that is returned from the API call.

MikeVedsted: I just mean that you could just use constants instead of importing useState and having all the setter functions as unused.

client/src/pages/ProfilePage/index.tsx
```diff
@@ -0,0 +1,86 @@
...
+    getUserInfo()
+  }, [userId])
+
+  const handleEditClick = () => {
+    history.push('/')
+  }
+
+  return (
+    <>
+      <Navbar />
```
MikeVedsted, line 41: we can just put navbar above the routes to avoid rerenders, so no need for it here.

MikeVedsted (approved):
Much better. There's are some serious issues with the responsivenes though.

Rostislav013 (approved):
Coolio

PR #113, opened 2020-12-15
[CHORE] Configure react-cookies for auth

MikeVedsted:
- [X] Implement cookieprovider
- [X] Store auth info in cookie
- [x] Use cookie to retrieve auth info
- [x] Refactor navbar to use info from cookie

### DIFF
--- a/client/src/components/Navbar/index.tsx
+++ b/client/src/components/Navbar/index.tsx
@@ -1,7 +1,6 @@
 import React, { useState } from 'react'
 import { Link } from 'react-router-dom'
 import { useCookies } from 'react-cookie'
-import jwt from 'jsonwebtoken'
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { faBell, faComment } from '@fortawesome/free-solid-svg-icons'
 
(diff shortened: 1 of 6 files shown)


client/src/components/Navbar/index.tsx
```diff
@@ -11,8 +10,7 @@ import './Navbar.scss'
 const Navbar = () => {
   const [navBg, setNavBg] = useState(true)
   const [cookies, setCookie] = useCookies(['x-auth-token'])
-  const decodedToken = jwt.decode(cookies['x-auth-token'])
```
Rostislav013, line 14: Do we still need here `'x-auth-token'` or can use `user`? Or does it really need smth in `useCookies()`?

Rostislav013 (approved):

Rostislav013:
If it works - it works.

Rostislav013:
Also can refactor `EventForm` components. It needs `user_id` too

client/src/components/Navbar/index.tsx
MikeVedsted: not sure how I missed this. It has been corrected now 👍

MikeVedsted:
> Also can refactor `EventForm` components. It needs `user_id` too

Now passing user_id as intended 👍

MarinaFroes (approved):

PR #115, opened 2020-12-15
[FEAT] Add route auth restriction

MikeVedsted:
Add component to limit route access to authenticated users.
Added implementation example to `src/Routes` an `App`


Rostislav013 (approved):

client/src/util/ProtectedRoute.tsx
```diff
@@ -1,24 +1,21 @@
 import React from 'react'
 import { Route, Redirect } from 'react-router-dom'
 import { ProtectedRouteProps, AppState } from '../types'
-import { useSelector } from 'react-redux'
```
Rostislav013, line 4: Fix imports

MikeVedsted: :shipit:

MarinaFroes (approved):
It seems good.

PR #117, opened 2020-12-15
[test] Add user service tests

Rostislav013:
Welcome to have a look at our page, dear reader :)

Feel free to propose tests to implement.


test/controllers/user.test.ts
```diff
@@ -7,18 +7,19 @@ import * as dbHelper from '../db/db-helper'
...
+    //await dbHelper.dropTempTables()
   })
 
   afterAll(async () => {
     // console.log(await (await db.query('SELECT * FROM userk;')).rows)
   })
 
-  it('should get all users', async () => {
+  // eslint-disable-next-line jest/no-disabled-tests
+  it.skip('should get all users', async () => {
```
MarinaFroes, line 22: Did you need to push commented code and skip this test for a specific reason?
You wouldn't need to disable the rule if you removed the `.skip`.

test/services/user.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    [
+      '40e6215d-b5c6-4896-987c-f3063678f608',
+      'Tester22',
+      'Testerov',
+      'testherethestthere@test.com',
+      '2017-03-18 08:21:36.175627+07'
+    ]
+  )
+  return await db.query('SELECT * FROM userk;')
+}
```
MarinaFroes, line 20: This function could also receive an optional user parameter for the cases when you want to create an extra user, like you did in line 66. We did something similar in our fullstack project.
I suggest to use arrow function here as well. `const createUser = async (override?: Partial<User>) => {}`

test/services/user.test.ts
```diff
@@ -0,0 +1,122 @@
...
+      'INSERT INTO userk (user_id, first_name, last_name, email, created_at) VALUES ($1, $2, $3, $4, $5);',
+      [
+        '40e6215d-b5c6-4896-987c-f3063678f608',
+        'Tester22',
+        'Testerov',
+        'testherethestthere@test.com',
+        '2017-03-18 08:21:36.175627+07'
+      ]
+    )
+    const user = await db.query('SELECT * FROM userk;')
```
MarinaFroes, line 46: Wouldn't it be the case to use the `createUser` function here?
And when checking if the user was created, shouldn't we use a user service to find the created user?

test/services/user.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    const user2 = await db.query(
+      'INSERT INTO userk (user_id, first_name, last_name, email, created_at) VALUES ($1, $2, $3, $4, $5);',
+      [
+        '40e6215d-b5c6-4896-987c-f3063678f508',
+        'Tester1',
+        'Testerov1',
+        'testherethestthereheh@test.com',
+        '2017-05-18 08:21:36.175627+07'
+      ]
+    )
```
MarinaFroes, line 75: As I mentioned before, if we pass a user param to `createUser`, you can use it here as well to create the second user.

test/services/user.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    const allUsers = await db.query('SELECT * FROM userk;')
+
+    expect(allUsers.rows.length).toEqual(2)
+  })
+
+  it('should not get a non-existing book', async () => {
+    const user = await createUser()
```
MarinaFroes, line 82: It really shouldn't get a **non-existing book** in our user table. 😄

test/services/user.test.ts
```diff
@@ -0,0 +1,122 @@
...
+      [
+        '40e6215d-b5c6-4896-987c-f3063678f508',
+        'Tester1',
+        'Testerov1',
+        'testherethestthereheh@test.com',
+        '2017-05-18 08:21:36.175627+07'
+      ]
+    )
+    const allUsers = await db.query('SELECT * FROM userk;')
+
```
MarinaFroes, line 77: Here we should test the `findAllUsers` endpoint, right?

MarinaFroes (changes requested):
Please, consider my comments below.

test/controllers/user.test.ts
Rostislav013: This one is for controllers still in progress. I would like to keep as it is now. 
Skip is temporary while doing other tests. Perfectly it should be without any skip.

Chiranjibichapagain (approved):
Looks good.

MikeVedsted (approved):

test/services/user.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    const findAllUsers = await db.query('SELECT * FROM userk;')
+
+    // Pay attention to row[number] where number is your needed row
+    expect(findAllUsers.rows.length).toEqual(2)
+    expect(user1.rows[0].first_name).toEqual('Rost')
+    expect(user2.rows[1].email).toEqual('Petrenko@dff.com')
+  })
+
+  it('should not get a non-existing event', async () => {
+    const user = await createUser()
```
MarinaFroes, line 72: 'should not get a non-existing _**user**_'

MarinaFroes (approved):
Great.

PR #137, opened 2020-12-16
[feat] Add footer component

Rostislav013:
Result:
Logos are changed to dark in footer.

footer

footerMOBILE

Feel free to suggest hot to improve BEM and styling.

On hover email is underlined.


Chiranjibichapagain (approved):
Hyvää!

client/src/components/Footer/index.tsx
```diff
@@ -0,0 +1,43 @@
+import React from 'react'
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+import { faGithub } from '@fortawesome/free-brands-svg-icons'
+
+import logo from '../../Assets/logoLight.svg'
```
MikeVedsted, line 5: I think you should use the other version of the logo for better contrast

client/src/components/Footer/Footer.scss
```diff
@@ -0,0 +1,41 @@
+@import '../../SCSS/config';
+
+.footer {
+  background-color: $default;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+
```
MikeVedsted, line 8: if your set `color: $textLight;` here, you don't have to do it again for other elements than the link.

client/src/components/Footer/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+          >
+            contact.joinme2020@gmail.com
+          </a>
+        </p>
+        <a
+          className='footer__link'
+          href='https://github.com/MikeVedsted/JoinMe'
+          target='a_blank'
+        >
+          <FontAwesomeIcon className='footer__icon' icon={faGithub} />
```
MikeVedsted, line 36: You can set the size already here, instead of styling your way to it. That way it is relative to the text it is in. 
`<FontAwesomeIcon icon={faGithub} size='2x'/>`

client/src/components/Footer/Footer.scss
```diff
@@ -0,0 +1,41 @@
+@import '../../SCSS/config';
+
+.footer {
+  background-color: $default;
```
MikeVedsted, line 4: your foot isn't sticky or absolutely placed. I guess that is not on purpose.

MikeVedsted (approved):
I know it is in the wireframe but maybe having the logo in the footer is a bit much.

client/src/components/Footer/Footer.scss
Rostislav013: Should it be sticky or absolutely placed?
Cuz I did it specially like this so it will be at the bottom of pages (usually) or can be modified later if some pages does not take full screen

client/src/components/Footer/index.tsx
Rostislav013: I used this way cuz it was alr used by Chiran this way. IF change here then need a change there. Let's keep this way. LAter when refactoring can change.

client/src/components/Footer/index.tsx
Rostislav013: image

Rostislav013: hmm

client/src/components/Footer/Footer.scss
MikeVedsted: We don't always know how much space a page's content will take up. 

Look at these examples for why that is an issue: 
fine
maybe not

PR #140, opened 2020-12-16
[feat] Add form slider component

### LINKED ISSUE #122: [FEAT] Add FormSlider component
Add a FormSlider component to the EventSearch component. 
This should include a slider to select the distance from the location specified by the user in the GoogleAutoComplete field. 

_Check the components division and naming in this figma file: https://www.figma.com/file/CZn7N02015QxKByVw9vRP1/JoinMe-Components?node-id=0%3A1_

DistanceSlider

This is the maximum distance I’m willing to travel to go to an event. 

**🏆 My dream solution would be**  
Just like Chiran’s layout on figma. 

**:2nd_place_medal: But I'd also consider it solved if**    
If it’s just a little bit similar to Chiran’s layout. 

**💭 If you were doing it, what would you do?**  
No idea. Google it.

**♻️ Additional context**  
It is required by the EventSearch component
It doesn’t require other components

**🚀 I'm ready for take off**  

MarinaFroes:
Fixes #122 
### Changes proposed in this pull request:
- This PR includes a FormSlider component (just frontend)

### When reviewing this PR, please consider:
~~- I had to use `useRef` for this and considering this should be avoided, let me know if there's a better way to accomplish the same thing.~~
~~- Maybe I should improve the naming for the classes ( I also get confused with elements and modifiers), if you have suggestions, please add comments.~~
- I had to change the position of the current value, otherwise, it would overlap the initial and final values of the range. Please check screenshots below. If this is a problem, we can think on another way to solve this.
- I also didn't know how to change the color only on the left of the track. 

#### Expectation:
Screenshot 2020-12-16 at 20 51 37

#### Reality (after fixing issue with thumb value positioning and background color):
_@MikeVedsted, thanks so much for the help in your review!_ 🥇 

Screenshot 2020-12-18 at 20 48 57

 _After merging this PR, I will add these changes to the `EventSearch` component in a new PR and it will look like this:_

EventSearch


client/src/App.tsx
```diff
@@ -1,43 +1,21 @@
...
-          }
+      <div style={{ width: '200px', padding: '20px' }}>
+        <FormSlider labelText='Distance from location' />
+      </div>
+      <div style={{ width: '600px', padding: '20px' }}>
+        <FormSlider
+          maxRange={250}
+          labelText='Distance from location'
+          defaultValue={50}
         />
```
Rostislav013, line 17: It will be removed before merging, right?

Rostislav013 (approved):

client/src/App.tsx
MarinaFroes: Yes, that's my intention. This is the working example in case you guys want to pull and test it.

client/src/components/FormSlider/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const [distance, setDistance] = useState(defaultValue)
+  const componentRef = useRef<HTMLDivElement>(null)
+
+  useEffect(() => {
+    if (componentRef.current) {
+      const width = componentRef.current.offsetWidth
+      componentRef.current.style.left = `calc(${distance} * (${
+        width / maxRange
+      }px))`
+    }
```
MikeVedsted, line 22: Since you ask for it in the PR... I think all the `useRef` is doing for you in this case is to set the left positioning as an inline style. As much as everyone dislikes inline styling maybe this is actually one of the acceptable use cases, as it would allow you to calculate and dynamically apply the margin or position needed to move the slider value.

client/src/components/FormSlider/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+import { FormSliderProps } from '../../types'
+import './FormSlider.scss'
+
+const FormSlider = ({
+  labelText,
+  minRange = 0,
+  maxRange = 100,
+  steps = 5,
+  defaultValue = 20
```
MikeVedsted, line 11: I really like your use of default values in this case :)

MikeVedsted (commented):
There's a slight issue with the placement of the current value. It wanders with the thumb, and is only centered at the default value. I'm willing to forget I saw that though 😅 


initial
right
left

MikeVedsted (commented):
> * Maybe I should improve the naming for the classes ( I also get confused with elements and modifiers), if you have suggestions, please add comments.

I think the "issue" with your naming here, is that you start of with an element at the top level. That restricts you a lot in your naming.

you could have a structure along these lines:

```
.slider {                                      // the overall component - previously form__slider
  ...
  &__range {                             // previously --range
    ...
    &::-webkit-slider-thumb {   // previously --range::-webkit-slider-thumb
      ...
    }
    &::-moz-range-thumb {     // previously --range::-moz-range-thumb 
      ...
    }
  }
  &__subset-labels {               // previously --subtitle-container
    ...
    &--text {                            // previously --subtitle
     ...
    }
  }
  &__value {                          // previously --value-container
    ...
    &--text {                          // previously --value
      ...
    }
  }
}
```

MikeVedsted:
> * I also didn't know how to change the color only on the left of the track.
First off,  I don't think this is too important. But if you want to implement it, I think inline styling (Again, I know, I'm sorry) is probably the easiest way to achieve it. 

You can style the `range` background as a gradient with four stops and update them dynamically. 
Gradient along these lines: 
`background: linear-gradient(90deg, $primary 0%,  $primary {currentValue},  $secondary {currentValue}, $secondary  100%);`

MarinaFroes:
> There's a slight issue with the placement of the current value. It wanders with the thumb, and is only centered at the default value. I'm willing to forget I saw that though 😅

 I know and that's killing me. I also need to change the color on the track to the left. I'll convert this to a draft, it's still work in progress. And I'll fix the rest of the issues as well.

Rostislav013 (approved):

client/src/components/FormSlider/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+  maxRange = 100,
+  steps = 5,
+  initialValue = 20
+}: FormSliderProps) => {
+  const [value, setValue] = useState(initialValue)
+  const [relativeValue, setRelativeValue] = useState(initialValue)
+  const [thumbPosition, setThumbPosition] = useState('')
+
+  useEffect(() => {
+    const relativeValue = ((value - minRange) * 100) / (maxRange - minRange)
```
MikeVedsted, line 17: Does it get confusing to have two constants with the same name (relativeValue )?

MikeVedsted (approved):
Nice job :) 
Only thing I'm missing, is a handling function that ties it together with form but that is being implemented when inserting in the form?

Chiranjibichapagain (approved):
:+1:

client/src/components/FormSlider/index.tsx
MarinaFroes: Good point, I'm gonna fix it.

PR #143, opened 2020-12-17
[FEAT] Add dropdown items

MikeVedsted:
Adding reusable dropdown item components for links and functions separately.

Related issues:
#142 and #114

Issue #114, opened 2020-12-15
[FEAT] Add logout


MikeVedsted:
**💡 I would really like to solve or include**   
Clearly and concisely describe the problem you are trying to solve.   

When logged in, the only way to log out is currently to close the browser entirely.
We should have and display a log out button when user is authenticated.

**👶 How would a user describe this?**   
Describe how users are affected by statements users might make or a user story.  
  
"How do I log out? Can anyone just use my device and info now?!"

**🏆 My dream solution would be**  
Describe the best possible scenario of this being implemented.  
Different config of google login button, that removes the cookie and redirects to landing page

**🚀 I'm ready for take off**  
Before submitting, please mark if you:  
- [x] Checked that this feature doesn't already exists  
- [x] Checked that a feature request doesn't already exists  
- [x] Went through the user flow, and understand the impact  
- [x] Made sure the request shows why it is important to users but doesn't exaggerate the value

MikeVedsted:
Need clarification on what the component will be embedded in

PR #144, opened 2020-12-17
[FEAT] Add nav dropdown container

MikeVedsted:
Adding dropdown container, styles and logout functionality to feed the item component. 

Related PR:
#143 
Related issues:
#141 and #114

### DIFF
--- a/client/src/components/NavDropdown/NavDropdown.scss
+++ b/client/src/components/NavDropdown/NavDropdown.scss
@@ -30,11 +30,11 @@
     align-items: center;
     width: 100%;
     padding: 10px 5px;
-  
+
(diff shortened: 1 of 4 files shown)


Rostislav013 (approved):

MarinaFroes (approved):
Great. 🏅 
For future improvement, do you think we could close it (hide it) when clicking anywhere outside of it?

And I think it's not supposed to go up when scrolling down, right (see screenshot)? I mean it probably would have to be fixed. But this can be a future improvement as well.
Screenshot 2020-12-17 at 18 59 59

Chiranjibichapagain (approved):
:+1:

PR #148, opened 2020-12-18
[feat] Add event search component

Chiranjibichapagain:
It is not much here. just collected dropdown input and button. The slider still needs to be added. We can refactor it later.


MikeVedsted (approved):
I guess as a placeholder, it is fine ;)

nannadao (approved):

PR #153, opened 2020-12-21
Fix button responsiveness

Chiranjibichapagain:


Rostislav013 (approved):

client/src/components/EventForm/index.tsx
```diff
@@ -8,7 +8,7 @@ import DropdownField from '../FormDropdownField'
 import GoogleAutoComplete from '../GoogleAutoComplete'
 import { useFormFields } from '../../hooks/useFormFields'
 import { eventCategories } from '../../util/constants/eventCategories'
-import './EventForm.scss'
+import './eventForm.scss'
```
MarinaFroes, line 12: Please, change it back to `./EventForm.scss` before merging it. 
The `Eventform.scss` is supposed to have a capital `E` in the name, if it doesn't it's because you don't have the most recent version in your machine. Could you please check that?

MarinaFroes (commented):
It's ok when testing it alone, but it's not working so fine in the `LandingPage` when I test it here. In viewports with width up to 1024px, it's larger, but when the viewport is 1025px, is narrow again.  But maybe we can merge it then we fix the `LandingPage` styles instead.

client/src/components/EventForm/index.tsx
Chiranjibichapagain: I pulled the latest changes and the scss file name is still eventForm.scss. So i will change the filename to EventChange

Chiranjibichapagain:
> It's ok when testing it alone, but it's not working so fine in the `LandingPage` when I test it here. In viewports with width up to 1024px, it's larger, but when the viewport is 1025px, is narrow again. But maybe we can merge it then we fix the `LandingPage` styles instead.

It is recommended to make larger buttons for mobile and tablets as the user will have to use their fingures. That's why i made them bigger in smaller screens. But we can agree on sizes later.

MarinaFroes:
> It is recommended to make larger buttons for mobile and tablets as the user will have to use their fingures. That's why i made them bigger in smaller screens. But we can agree on sizes later.

I can show what I mean in the meeting tomorrow. But it's ok for now.

MarinaFroes (approved):

PR #156, opened 2020-12-21
[feat] Add slider to EventSearch component

### LINKED ISSUE #123: [FEAT] Add slider to EventSearch component
Add an EventSearch component to the Event component. 
This should include a FormDropdownField for the category of event, a GoogleAutoComplete field for the location of the event and a DistanceSlider for the distance the user is willing to travel to go to the event. 

_Check the components division and naming in this figma file: https://www.figma.com/file/CZn7N02015QxKByVw9vRP1/JoinMe-Components?node-id=0%3A1_

EventSearch

This is where I select the options to filter the events I’m interested on. 

**🏆 My dream solution would be**  
Just like Chiran’s layout on figma. 

**:2nd_place_medal: But I'd also consider it solved if**    
If it’s just a little bit similar to Chiran’s layout. 

**💭 If you were doing it, what would you do?**  
- Import all the required components

**♻️ Additional context**  
It is required by the Event component
It requires the FormDropdownField, GoogleAutoComplete and the DistanceSlider components

**🚀 I'm ready for take off**  

MarinaFroes:
Fixes #123 

Guys, I added the slider component to the `EventSearch` and in the `App.tsx` (_working example, but it will be removed before merging_) I can get the data from the `category`, `address` and `distance`, but I'm not sure this is the best solution. Could you please check this and let me know if I need to fix something?
@Chiranjibichapagain  is working on the `HomePage` so he would import the `EventSearch` there, filter the events based on the data that he'll get and then return the filtered events.

It looks like this right now, but we can adjust it later, the most important is to check if the logic is correct.
Screenshot 2020-12-21 at 14 46 56

### DIFF
--- a/client/src/components/EventSearch/index.tsx
+++ b/client/src/components/EventSearch/index.tsx
@@ -9,8 +9,8 @@ import Button from '../Button'
 import './EventSearch.scss'
 
 const EventSearch = ({
-  fields,
-  handleFields,
+  distance,
+  handleFieldChange,
   handleSubmit,
   setAddress
 }: EventSearchProps) => {
@@ -21,7 +21,7 @@ const EventSearch = ({
         label='Category'
         id='category'
         options={eventCategories}
-        onBlur={handleFields}
+        onBlur={handleFieldChange}
       />
       <label className='form__label'>
         Location
@@ -31,8 +31,8 @@ const EventSearch = ({
         Distance from location
         <FormSlider
           id='distance'
-          value={fields.distance}
-          onChange={handleFields}
+          value={distance}
+          onChange={handleFieldChange}
         />
       </label>
       <div className='search-box__button'>
--- a/client/src/components/EventSearch/index.tsx
+++ b/client/src/components/EventSearch/index.tsx
@@ -2,26 +2,45 @@ import React from 'react'
 
 import FormDropdownField from '../FormDropdownField'
 import GoogleAutoComplete from '../GoogleAutoComplete'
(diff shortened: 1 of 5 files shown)


Chiranjibichapagain (approved):
:+1:

client/src/components/EventSearch/index.tsx
```diff
@@ -2,26 +2,45 @@ import React from 'react'
...
-        onBlur={() => console.log('blurred!!')}
+        onBlur={handleFields}
       />
-      <GoogleAutoComplete handleAddress={() => console.log('handled!!')} />
+      <label className='form__label'>
+        Location
+        <GoogleAutoComplete handleAddress={setAddress} />
+      </label>
+      <label className='form__label'>
+        Distance from location
```
MikeVedsted, line 31: You could move the label to the slider component, like for the input fields. This is fine for now though, but maybe @Chiranjibichapagain could throw it in there? :)

MikeVedsted (approved):
Chiran said he would fix the background color for the input when finishing this component, so good for now 👍

client/src/components/EventSearch/index.tsx
MarinaFroes: I removed the label because I was having some issues with it when styling the component. But if he manages to fix it and include it again, it would be great.

MarinaFroes:
> Chiran said he would fix the background color for the input when finishing this component, so good for now 👍

It definitely needs styles adjustments. 😄

Issue #62, opened 2020-12-03
[FEAT] Edit user information page


nannadao:
**💡 I would really like to solve or include**   
Add a page where we can edit user information
User should be able to update their first name, last name, add interest, profile intro, and upload profile picture. User **SHOULD NOT** be able to change the email.

_Check the components division and naming in this figma file: https://www.figma.com/file/CZn7N02015QxKByVw9vRP1/JoinMe-Components?node-id=0%3A1_

![UserDetailsPage](https://user-images.githubusercontent.com/45233290/102264795-6918de80-3f16-11eb-8606-1e7c13434f4c.png)

**👶 How would a user describe this?**   
In this page I can edit my own information

**🏆 My dream solution would be**  
Just like Chiran’s layout on figma. 

**:2nd_place_medal: But I'd also consider it solved if**    
If it’s just a little bit similar to Chiran’s layout. 

**💭 If you were doing it, what would you do?**  
- Follow figma file

**♻️ Additional context**  
It requires Navbar, Footer and UserForm

**🚀 I'm ready for take off**  
Before submitting, please mark if you:  
- [x] Checked that this feature doesn't already exists 
- [x] Checked that a feature request doesn't already exists  
- [x] Went through the user flow, and understand the impact  
- [x] Made sure the request shows why it is important to users but doesn't exaggerate the value

PR #157, opened 2020-12-21
[FEAT] Add edit user for front end

MikeVedsted:
- [x] Added profile form 
- [x] added reusable page for both inital account set up and user requested updates

Related issue:
#62


Rostislav013 (approved):

client/src/App.tsx
```diff
@@ -1,5 +1,5 @@
-import React, { useState } from 'react'
-// import { Link } from 'react-router-dom'
+
+import React from 'react'
```
MarinaFroes, line 2: Didn't you get a prettier error with this empty line here?

client/src/components/AccountForm/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+  const [fields, handleFieldChange] = useFormFields({
+    email: '',
+    first_name: '',
+    last_name: '',
+    date_of_birth: '',
+    gender: '',
+    base_address: '',
+    profile_text: '',
+    profile_image: ''
+  })
```
MarinaFroes, line 25: I think the form could come pre-filled with the data we have from Google. What do you think? Then of course the user would be able to update it, but it would be necessary less input from their side if they only need to add the fields that google doesn't have.

client/src/components/AccountForm/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+          id='profile_image'
+          label='Image url'
+          value={fields.profile_image}
+          onChange={handleFieldChange}
+        />
+        <img
+          className='image-upload__image'
+          src={fields.profile_image}
+          alt='What we currently display'
+        />
```
MarinaFroes, line 109: We can't see the image yet, but it's because we're not passing google info here yet, right?

MarinaFroes (approved):
It seems good, nice work.
But it's not actually working yet, right? Because there's another PR to refactor update user service and I tried it here but I could not update my data.

MarinaFroes:
> But it's not actually working yet, right? Because there's another PR to refactor update user service and I tried it here but I could not update my data.

It didn't work because I was not authenticated and the error handling in the `isAuthenticated` middleware was not working properly, so I couldn't see any error message. 😬

client/src/App.tsx
MikeVedsted: I think this happened when resolving merge conflicts with the browser tool. I get the error as well in VSCode and will push a change now.

client/src/components/AccountForm/index.tsx
MikeVedsted: I actually had that to begin with but it was causing some issues, so I simplified it away.
I agree it is a better experience, so I will try to find a better implementation.

client/src/components/AccountForm/index.tsx
MikeVedsted: It will only display what is in the fields at that moment. Since fields are currently empty by default, it doesn't display anything on the first render. Try adding an image url in the initial state to see it in action :) (and update the string in the form!)

MikeVedsted:
> It seems good, nice work.
> But it's not actually working yet, right? Because there's another PR to refactor update user service and I tried it here but I could not update my data.


Back end changes are set to merge into this branch, so until then, this is only for the front end. 

> It didn't work because I was not authenticated and the error handling in the `isAuthenticated` middleware was not working properly, so I couldn't see any error message. 😬

It should work to the extend that the old queries allowed for. It currently won't update all fields. As far as I remember without checking, it isn't updating address, gender and date of birth

client/src/components/AccountForm/index.tsx
MarinaFroes: Ok, but it doesn't need to be done right now.

client/src/components/AccountForm/index.tsx
MarinaFroes: I did that. 😄  If we don't get data from google in this current implementation, what about using a placeholder image (dummy link like this one: https://dummyimage.com/200x200/aaaaaa/ffffff&text=Add+your+photo+here) or even not showing the image element if there's no url? Right now we only see a broken image.

MikeVedsted: Good point. Setting a dummy in the field will submit it, if the user doesn't change it, so I prefer the conditional rendering option. 
I've just pushed that

PR #160, opened 2020-12-22
[feat] User Profile Page

nannadao:
Close issue #84


client/src/pages/ProfilePage/ProfilePage.scss
```diff
@@ -0,0 +1,104 @@
+@import '../../SCSS/config';
+
+.profile {
+  padding-top: 95px;
+  margin: 0 2.5em;
+  justify-content: stretch;
+  @include flexStart;
```
MikeVedsted, line 7: Looks like you forgot to submit your edits to the `_config` for scss.
There's no @flexStart or $lineDim on this branch

client/src/pages/ProfilePage/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    number: '',
+    postal_code: '',
+    city: '',
+    country: ''
+  })
+
+  useEffect(() => {
+    async function getUserInfo() {
+      try {
+        const { data } = await axios.get(`/api/v1/users/${userId}`)
```
MikeVedsted, line 32: nice destructuring! We should propably do this on the other requests :)

client/src/pages/ProfilePage/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+import { useHistory, useParams } from 'react-router-dom'
+
+import Button from '../../components/Button'
+import Navbar from '../../components/Navbar'
+import { ProfilePageParamProps } from '../../types'
+import './ProfilePage.scss'
+
+const ProfilePage = () => {
+  const history = useHistory()
+  const { userId } = useParams<ProfilePageParamProps>()
```
MikeVedsted, line 12: everywhere else, we are getting the id from the cookies, so it would probably be better to keep it the same here.

client/src/pages/ProfilePage/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    }
+    getUserInfo()
+  }, [userId])
+
+  const handleEditClick = () => {
+    history.push('/')
+  }
+
+  return (
+    <>
```
MikeVedsted, line 46: No need to wrap the return in a fragment, when it is all inside one div

client/src/pages/ProfilePage/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+              <p className='profile__details--text'>
+                `${userInfo.street} ${userInfo.number}, ${userInfo.postal_code}{' '}
+                ${userInfo.city}, ${userInfo.country}`
+              </p>
+            ) : (
+              <p className='profile__details--text'> - </p>
+            )}
+          </div>
+        </div>
+      </div>
```
MikeVedsted, line 105: This could be DRY'ed up by making a few smaller components but as with some of the other pages, we can refactor later.

client/src/pages/ProfilePage/index.tsx
```diff
@@ -0,0 +1,110 @@
+import React, { useEffect, useState } from 'react'
+import axios from 'axios'
+import { useHistory, useParams } from 'react-router-dom'
+
+import Button from '../../components/Button'
+import Navbar from '../../components/Navbar'
```
MikeVedsted, line 6: navbar can be removed from imports

MikeVedsted (approved):
Looks good!

in terms of naming, I don't see the page styled with a sidebar but rather, what is names sidebar is a banner. Maybe your styling is of, otherwise please rename to a more appropriate name.

client/src/pages/ProfilePage/index.tsx
Rostislav013: But other users also can check other users’ pages

nannadao: Yes, I guess this page is accessible for everyone so getting the ID from cookies would mean that one can only see theirs profile page

MikeVedsted: I think we agreed on having a conditional render at some point, right? 
Any authenticated user can access the page but only the 'owner' should see sensitive information, right? 

I understood this page as being the current users page, in which case getting the id from cookies would be sensible. But if we are doing what I just described, it makes sense to leave it as is, and check if the id in cookies is the same as the one in params.

Sorry for the confusion

client/src/pages/ProfilePage/ProfilePage.scss
```diff
@@ -0,0 +1,104 @@
...
+    width: 30vw;
+  }
+}
+
+@media only screen and (max-width: 765px) {
+  .profile__header {
+    > * {
+      width: calc(100% / 3);
+    }
+    &--text {
```
Rostislav013, line 90: A space between

client/src/pages/ProfilePage/ProfilePage.scss
```diff
@@ -0,0 +1,104 @@
...
+      font-size: 20px;
+      text-align: center;
+    }
+    &--divider-optional {
```
Rostislav013, line 94: A space between

Rostislav013 (approved):

PR #167, opened 2020-12-22
[CHORE] Code review pt2

MikeVedsted:
Todos that I could easily fix have been included.
Issues have been created for todos that require investigation or are more than quick fixes


client/src/pages/LandingPage/index.tsx
```diff
@@ -9,70 +8,64 @@ import { eventCategories } from '../../util/constants/eventCategories'
...
       const response = await axios({
         method: 'GET',
-        url: 'http://localhost:5000/api/v1/users'
+        url: '/api/v1/users'
       })
       const data = response.data.length
       setUsers(data)
     } catch (error) {
-      setError(error.message)
+      setUsers('Many')
```
Rostislav013, line 27: Is it good idea to get all users here? 
And do we need it?

MikeVedsted: Well, if we access to that route in the backend, it won't work... We could make a route to just return the user count maybe?

Rostislav013: Just any random word(number) heh

MarinaFroes (approved):
I think it's ok.

PR #169, opened 2021-01-01
[Fix] Fix landing page responsiveness

Chiranjibichapagain:


client/src/pages/LandingPage/LandingPage.scss
```diff
@@ -53,66 +53,96 @@
...
       &--right {
-        top: 60vh;
-        left: 50vw;
+        left: 47vw;
+      }
+    }
+
+    &__text {
+      width: 75%;
+      &--highlight {
```
Rostislav013, line 104: Probably here need a space too like in line 92.

Rostislav013 (approved):

client/src/components/Navbar/Navbar.scss
```diff
@@ -19,7 +19,7 @@
   &__image {
     &--logo {
       height: 60px;
-      margin-left: 50px;
+      margin-left: 3vw;
```
MikeVedsted, line 22: this change requires that we make the same change on the MyEvents page, where the margin-left of the sidebar is set to the same value, as far as I remember.

MikeVedsted (approved):

PR #178, opened 2021-01-04
[feat] Add comment component

MikeVedsted:
Adding comment component to replace placeholder in event and unblock  comment section. 
image


Rostislav013 (approved):

client/src/components/EventComment/EventComment.scss
```diff
@@ -0,0 +1,39 @@
+@import '../../SCSS/config';
+
+.comment {
+  width: 100%;
+  background-color: $backgroundDim;
+  margin: 10px 0;
+  display: flex;
+  border-radius: 8px;
+
+  &__image {
```
Chiranjibichapagain, line 10: Maybe the height of the profile image can be smaller, looks too tall?

Chiranjibichapagain (approved):
:+1:

client/src/components/EventComment/EventComment.scss
MikeVedsted: In the example, I am using an image in a slightly taller image than usual, so I think it looks worse than it generally would. I've resized to `max-height: 40px` though, as that should also be enough :)

PR #183, opened 2021-01-04
[fix] Add redirection when user is logged in

Rostislav013:
So now `/` is for events. If user not logged in, user is redirected to `/get-started` where user logs in. If/when user logged in - user is redirected to `/` to see all events.


client/src/pages/LandingPage/index.tsx
```diff
@@ -28,6 +33,11 @@ const LandingPage = () => {
     }
   }
 
+  const handleGetStartedButton = () => {
+    console.log('hi')
```
MikeVedsted, line 37: please remove the console.log

MikeVedsted (approved):
It would be better if we also update the log in function, to redirect users after log in. There is no need for users to go to this page at all if they are authenticated

Chiranjibichapagain (approved):
:+1:

client/src/Routes.tsx
```diff
@@ -10,14 +10,13 @@ import MyEventsPage from './pages/MyEventsPage'
...
-    <Route exact path='/' component={LandingPage} />
+    <Route exact path='/' component={HomePage} />
     <Route exact path='/user/:userId' component={ProfilePage} />
-    <ProtectedRoute exact path='/home' component={HomePage} />
+    <Route exact path='/get-started' component={LandingPage} />
     <ProtectedRoute
       path='/user/:userId/:contentName'
       component={MyEventsPage}
     />
-    <Route path='/' component={LandingPage} />
```
MikeVedsted, line 20: We should have something for  this path (line 20) to catch 404s. It needs to be at the bottom and without the `exact` prop, to catch all paths that do not have a match. For now, please add the `HomePage` since it is should be a protected path that will redirect to `LandingPage` if user is not authorized.

client/src/Routes.tsx
```diff
@@ -10,14 +10,13 @@ import MyEventsPage from './pages/MyEventsPage'
 
 const Routes = () => (
   <Switch>
-    <Route exact path='/' component={LandingPage} />
+    <Route exact path='/' component={HomePage} />
     <Route exact path='/user/:userId' component={ProfilePage} />
-    <ProtectedRoute exact path='/home' component={HomePage} />
+    <Route exact path='/get-started' component={LandingPage} />
```
MikeVedsted, line 15: Please make the `HomePage` a protected path. Non-authorized users should not be able to access this page.

client/src/pages/Homepage/index.tsx
```diff
@@ -11,8 +13,14 @@ import './Homepage.scss'
...
   const [address, setAddress] = useState({})
-
   const [isModalOpen, setIsModalOpen] = useState(false)
+  const history = useHistory()
+  const [cookies] = useCookies(['user'])
+  const { user_id } = cookies.user || ''
+
+  useEffect(() => {
+    !user_id && history.push('/get-started')
+  })
```
MikeVedsted, line 23: No need for this if we keep `HomePage` as a protected route

MikeVedsted (changes requested):
Please update the redirect in ProtectedRoute to '/get-started'.

client/src/pages/Homepage/index.tsx
Rostislav013: It depends. Currently protected route redirects to `/`. `HomePage` itself is on `/` path.

I will update protected route then to redirect to `/get-started`.

MikeVedsted: Hence my other comments :) Sorry if that was confusing

client/src/Routes.tsx
Rostislav013: Sounds like a new issue

MikeVedsted: No need, it is just what you have deleted here :D

Rostislav013: `HomePage` and `LandingPage` are both there. `HomePage` is in protected route.

PR #189, opened 2021-01-05
[FEAT] Add comment section component

MikeVedsted:
Done: 
- [x] Create comment section wrapper
- [x] Modify Event component and HomePage page to pass needed props
- [x] Remove styling from Event component that is no longer needed
- [x] Create hook for fetching EventComments

### DIFF
--- a/client/src/hooks/useEventComments.tsx
+++ b/client/src/hooks/useEventComments.tsx
@@ -5,7 +5,6 @@ import { EventId, Comment } from '../types'
 
 const useEventComments = (eventId: EventId) => {
   const [comments, setComments] = useState<Comment[]>([])
-  const [error, setError] = useState()
 
   useEffect(() => {
     fetchComments()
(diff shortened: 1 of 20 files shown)


client/src/hooks/useEventComments.tsx
```diff
@@ -0,0 +1,29 @@
...
+  useEffect(() => {
+    fetchComments()
+  }, [])
+
+  const fetchComments = async () => {
+    try {
+      const response = await axios({
+        method: 'GET',
+        url: `/api/v1/comments/${eventId}`
+      })
```
Rostislav013, line 19: Why not  using just:  `const response = await axios.get(`/api/v1/comments/${eventId}`)`?
Optimization :)

Rostislav013 (approved):

client/src/hooks/useEventComments.tsx
MikeVedsted: This PR was set as a draft, so there was no need for review yet - but thanks anyway :) 

I chose that style, to be aligned with how we do it in other places. But yea.. I've changed it to a `.get()` function instead in the now ready version

PR #190, opened 2021-01-05
[Feat] Add participant component

Chiranjibichapagain:
participant


client/src/components/EventParticipant/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+        <p className='participant__name'>{`${user.first_name} ${user.last_name}`}</p>
+      </Link>
+      <FontAwesomeIcon
+        onClick={handleDelete}
+        className='participant__delete'
+        icon={'trash'}
+      />
+    </div>
+  )
+}
```
MikeVedsted, line 26: Please destructure the user object as per the guidelines and use those values instead

client/src/components/EventParticipant/index.tsx
```diff
@@ -0,0 +1,28 @@
...
+import { EventParticipantProps } from '../../types'
+import './EventParticipant.scss'
+
+const EventParticipant = ({ user, handleDelete }: EventParticipantProps) => {
+  return (
+    <div className='participant'>
+      <img
+        className='participant__image'
+        src={user.profile_image}
+        alt='profile'
```
MikeVedsted, line 14: it would be better to use the users name as the alt, as profile isn't very informative

client/src/components/EventParticipant/EventParticipant.scss
```diff
@@ -0,0 +1,33 @@
...
+  margin: 10px 15px;
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
+
+  &__image {
+    width: 30px;
+    object-fit: cover;
+    border-radius: 4px;
+  }
```
MikeVedsted, line 14: I think it would be better to set both the height and width of the image, as the height will dictate the size of the wrapping div. That means that the rows won't be equally tall when mapped, in case the user has a taller profile image.

client/src/components/EventParticipant/EventParticipant.scss
```diff
@@ -0,0 +1,33 @@
...
+
+  &__name {
+    width: 200px;
+    text-align: left;
+    color: $link;
+    font-family: $nunito;
+    font-size: $text;
+    font-weight: bold;
+  }
```
MikeVedsted, line 27: By defining the width with pixels, aren't we risking it looking weird for long names? Have you tried with your own full name?

MikeVedsted (approved):

Rostislav013 (approved):

Rostislav013:
Remember to fetch updates from develop before merging

client/src/components/EventParticipant/EventParticipant.scss
Chiranjibichapagain: My name fits. but gonna increase to fit 5 characters longer than my name :D Doing this to make delete btn aligned.

PR #194, opened 2021-01-05
[FEAT] Add comment crud operations to BE

MikeVedsted:


src/routers/comment.ts
```diff
@@ -0,0 +1,18 @@
...
+  updateComment,
+  deleteComment
+} from '../controllers/comment'
+
+const router = express.Router()
+
+router.post('/:eventId', createComment)
+router.get('/:eventId', findCommentsByEventId)
+router.put('/:commentId', updateComment)
+router.delete('/:commentId', deleteComment)
```
Rostislav013, line 16: Do we need to protect some routes? Only owner can edit, delete for ex

src/services/comment.ts
```diff
@@ -0,0 +1,75 @@
...
+    const query = `
+      SELECT * FROM comment
+      LEFT JOIN userk ON comment.userk = userk.user_id
+      WHERE event = $1
+      ORDER BY commented_at
+    `
+    const DBResponse = await db.query(query, [eventId])
+    const comments: Comment[] = DBResponse.rows
+
+    return comments
```
Rostislav013, line 27: Are they sorted by date by default or sorting by date/time is needed?

src/routers/comment.ts
MikeVedsted: Yes. There seems to be an issue with the protection though, so we will have to double check that first, and maybe protect them later.

src/services/comment.ts
MikeVedsted: That depends on the DB. I'm not sure we have changed the default sorting, so if you think it is needed here, it would be better to open a new issue. 
When working on the update operation, it didn't seem like they were sorted by the date.

Rostislav013: Or it can be sorted in frontend ha

Rostislav013 (approved):

Chiranjibichapagain (approved):
:+1:

PR #198, opened 2021-01-05
[Feat] Add event manage dropdown

Chiranjibichapagain:
manage


client/src/types.ts
```diff
@@ -1,4 +1,5 @@
 import { IconProp } from '@fortawesome/fontawesome-svg-core'
+import { Dispatch, SetStateAction } from 'react'
```
Rostislav013, line 2: Looks like we don't need line 2 here and can delete it?

Rostislav013 (approved):

Rostislav013:
You are a gem.
Remember to fetch updates before merging.

client/src/types.ts
MikeVedsted: it's needed for the type EventSearchProps but I don't know why. I think that is a mistake but as long as that type uses it, the import is needed.

MikeVedsted (approved):

PR #201, opened 2021-01-05
[feat] Close login modal on success

Rostislav013:
It is probably not needed, but I just wanted to close the issue with PR.


MikeVedsted (approved):
Not exactly what was meant but since we changed the routing this works even better 😄 👍

Chiranjibichapagain (approved):
:+1:

Rostislav013:
> Not exactly what was meant but since we changed the routing this works even better

Why not exactly? One of the solution in issue was is to close the modal oh oh

PR #202, opened 2021-01-05
[refactor] Move icon library to its own file

Rostislav013:


MikeVedsted (approved):
Just se! 
Would it make sense to move it to helpers or up one level though, as it is not a constant file per se?

Rostislav013:
> Just se!
> Would it make sense to move it to helpers or up one level though, as it is not a constant file per se?

Moved one lvl but not sure up or not heh

MikeVedsted:
> Moved one lvl but not sure up or not heh
You did exactly what I meant, so perfect from my end

Chiranjibichapagain (approved):
:+1:

PR #204, opened 2021-01-05
[refactor] Remove user request in landing page

Rostislav013:
Deleted user request cuz this route will be protected in backend, so it will not work plus I am not sure if we will allow users to see list of all users.


MikeVedsted:
Nothing wrong with this. But check out my solution in #210 - I think this is a better solution, if we want to keep the count while securing user data

MikeVedsted (approved):

Chiranjibichapagain (approved):

Rostislav013:
I am gonna merge this one. 

#210 can merge tomorrow after meeting.

PR #205, opened 2021-01-05
[feat] Protect user routes in backend

Rostislav013:
**Achtung. Your opinion is needed.**

I have protected some user routes.

<del>1. I need your opinion about `findAllUsers`. I think we don't need that. What do u think? It could be used only for a superUser.
 **DECIDED** to delete and get it back if and when needed.</del>
 <del>2.When I deleted `findAllUsers` route, Events components is a bit broken, It doesn't show names of event creators cuz of the 
    way it was used: find all users then using method `find` - find needed user.  </del>
  
<del>I think this way does not suit us in case not having `findAllUsers` service plus what if we have 2 millions users. Will be lagging. 
    So we need to refactor `**Event component**`.
    `/components/Event/index.tsx`
 </del>
3. I rmb we had some agreement about get info about user. If it is onwer show more on frontend, if not - less. But in this way we 
    are still passing all data about the user to frontend. Which isnt so good. 
    **DECIDED** to make a new route.
   
   To do:
- [ ] Fix requests on frontend (for ex pass additional data)


MikeVedsted (approved):
> 1. I need your opinion about `findAllUsers`. I think we don't need that. What do u think? It could be used only for a superUser.
>    **DECIDED** to delete and get it back if and when needed.

This is fine by me, we can add the route back in when/if we create admin functions.

> 2. When I deleted `findAllUsers` route, Events components ... doesn't show names of event creators ...  So we need to refactor `**Event component**`.

Yup. We discussed this already when Chiran created the function. It's a bad idea to expose all the data, which we are otherwise trying to keep confidential. The options are to return data depending on ownership or have separate routes. I believe we agreed to have two separate routes.

... 

Could you please do the same for the event routers while your are at it?

src/routers/user.ts
```diff
@@ -1,20 +1,14 @@
...
-} from '../controllers/user'
+import { isAuthenticated, isOwner } from '../middlewares/authentication'
+import { findUserById, updateUser, googleLogin, deleteUser } from '../controllers/user'
 
 const router = express.Router()
 
-router.get('/', findAllUsers)
-router.get('/:userId', findUserById)
+router.get('/:userId', isAuthenticated, isOwner, findUserById)
+// FIX create issue: Add one more route for limited user info (not owner)
```
MikeVedsted, line 9: Please created the relevant issue and remove comment

MikeVedsted (approved):

Chiranjibichapagain (approved):
:+1:

PR #207, opened 2021-01-06
[refactor] Edit user profile page

Rostislav013:
**Attention. Your opinion is required.**

<del>So we got a page for Account set up. It seems like it's the first one after creating an account. 
But what about just to edit user page at any time? If we use same set up account, it might a bit strange to get `For a better experience finding events, let's get your profile set up!` message every time.</del>

<del>Current flow: There is a page `AccountSetup` where is called component `AccountForm` to edit data.
What I want is to create a page `AccountForm`. **What do u think about that?**</del>

<del>**DECIDED** to make a new page for edit profile using AccountForm component. </del>

Please, pay more attention to type for `user, userState` n so on in `store`.

To do:
- [X] Check/fix updating user data

<del>I couldn't find a nicer way to implement it. Check `services/user` file `updateUser` method, let me know if u know a better way to implement it. Main idea is that if address isn't in update I need to ignore it n update the user without address.</del>


Chiranjibichapagain:
Profile edit page and profile setup page should be different. I mean a
little different. Profile setup page can be visited only once by a user in
the beginning, which includes the message 'For a better..'  Profile edit
page won't have any message in it just can alter the information there.
Pages/ component names can be ProfileSetup and ProfileEdit.

Rostislav013:
I will make then a file page/ProfileEdit. Where I will call AccountForm component. How does it sound? Is that what u meant?

MikeVedsted:
Make sure to reflect the latest changes to routes before undrafting this PR.

MikeVedsted:
@Rostislav013 check out the changes to address handling in #247 - I think you are dealing with the same issue

Chiranjibichapagain (approved):
:+1:

client/src/pages/ProfileEdit/ProfileEdit.scss
```diff
@@ -0,0 +1,28 @@
+@import '../../SCSS/config';
+
+.page {
+  min-height: 100vh;
```
MikeVedsted, line 4: after the changes to the footer position, min-height shouldn't be necessary anymore

client/src/pages/ProfileEdit/ProfileEdit.scss
```diff
@@ -0,0 +1,28 @@
...
+  &__field {
+    display: block;
+  }
+
+  &__image {
+    width: 150px;
+    margin-left: 10px;
+    border-radius: 16px;
+  }
+}
```
MikeVedsted, line 28: make sure this isn't repeated in both the component and the page

client/src/pages/ProfileEdit/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+  // FIX? Add loading component/icon anything if needed
+  // if (loading) return <div>LOADING</div>
+  return (
+    !loading && (
+      <div className='page'>
+        <h2 className='page__title'>
+          Hi {first_name}! Let&apos;s make some changes.
+        </h2>
+        {userId === user_id && <AccountForm userId={user_id} />}
+      </div>
```
MikeVedsted, line 42: why is this check necessary?

client/src/pages/ProfileEdit/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+        setLoading(true)
+        const { data } = await axios.get(`/api/v1/users/${user_id}`)
+        dispatch(addUser(data))
+        setLoading(false)
+      } catch (error) {
+        console.log(error)
+      }
+    }
+    getUserInfo()
+  }, [dispatch, user_id])
```
MikeVedsted, line 31: We are generally using hooks for this

client/src/pages/ProfileEdit/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+  // FIX? Add loading component/icon anything if needed
+  // if (loading) return <div>LOADING</div>
+  return (
+    !loading && (
```
MikeVedsted, line 36: We could implement a loading component. But please create an issue for it and remove the comment/implementation here, as it would need to be done in several places at the same time.

src/services/user.ts
```diff
@@ -77,52 +77,65 @@ const updateUser = async (userId: string, update: Partial<User>) => {
       last_name = user.last_name,
```
MikeVedsted, line 77: Please add a message to the error on line 72

src/services/user.ts
```diff
@@ -77,52 +77,65 @@ const updateUser = async (userId: string, update: Partial<User>) => {
...
+    const updateUser = await db.query(updateUserQueryWithoutAddress, [
       userId,
       first_name,
       last_name,
       profile_image,
       profile_text,
-      addressId,
       date_of_birth,
       gender
     ])
```
MikeVedsted, line 141: It is not necessary to have two separate queries like this. Please follow the same logic as for updateEvent

MikeVedsted (changes requested):
for some reason, after updating address, birthday and gender the new values are not displayed when going back to the form. I think this might be because of the way you are using redux.  The address components aren't being updated in the state.

In general, if you are choosing to use redux, we should start refactoring to inject it everywhere.

client/src/pages/ProfileEdit/index.tsx
Rostislav013: when they dont much user won't see a form. 
I fetch user data using user's id from cookies. Might be situation when userId is different from path but the page still shows correct user's data.

client/src/components/AccountForm/index.tsx
```diff
@@ -81,8 +92,12 @@ const AccountForm = ({ userId }: AccountFormProps) => {
...
-        value={'Saved address: ' + fields.base_address}
+        label='Address'
+        value={
+          user[0].street
+            ? 'Saved address: ' + userAddress
+            : 'Saved address: No address saved'
+        }
         readOnly
       />
       <GoogleAutoComplete handleAddress={setAddress} />
```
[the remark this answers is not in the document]
Rostislav013: Line 122 is different compare what is here. Specify again what do u mean

client/src/pages/ProfileEdit/index.tsx
Rostislav013: Didn't you created it that issue [issue249](https://github.com/MikeVedsted/JoinMe/issues/249) ?

client/src/components/AccountForm/index.tsx
```diff
@@ -81,8 +92,12 @@ const AccountForm = ({ userId }: AccountFormProps) => {
...
-        label='Default start address'
-        value={'Saved address: ' + fields.base_address}
+        label='Address'
+        value={
+          user[0].street
+            ? 'Saved address: ' + userAddress
+            : 'Saved address: No address saved'
+        }
         readOnly
       />
```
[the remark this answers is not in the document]
Rostislav013: Which PR/component?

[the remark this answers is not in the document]
Rostislav013: Used one from [PR247](https://github.com/MikeVedsted/JoinMe/pull/247/files), let me know if u meant this one

Rostislav013:
> for some reason, after updating address, birthday and gender the new values are not displayed when going back to the form. I think this might be because of the way you are using redux. The address components aren't being updated in the state.
> 
> In general, if you are choosing to use redux, we should start refactoring to inject it everywhere.

I think problem was with not displaying it in inputs. Check again, I added displaying correctly

client/src/components/AccountForm/index.tsx
```diff
@@ -8,42 +9,53 @@ import FormDropdownField from '../FormDropdownField'
...
+      setAddress({
+        street: user[0].street,
+        number: user[0].number,
+        postal_code: user[0].postal_code,
+        city: user[0].city,
+        country: user[0].country,
+        lat: user[0].lat,
+        lng: user[0].lng
+      })
+  }, [user])
```
MikeVedsted, line 40: if you change the user state to being a user object instead of an array, you won't have to always select [0]. This also makes sense if you think about that you should always _only_ have a single user's data available.

src/services/user.ts
```diff
@@ -67,55 +67,50 @@ const updateUser = async (userId: string, update: Partial<User>) => {
   try {
     const userResponse = await db.query('SELECT * FROM userk WHERE user_id = $1', [userId])
     const user: User = userResponse.rows[0]
-
+    console.log(user)
     if (!user) {
```
MikeVedsted, line 71: Please remove console.log

src/services/user.ts
```diff
@@ -67,55 +67,50 @@ const updateUser = async (userId: string, update: Partial<User>) => {
...
-      [lat, lng]
-    )
-
-    if (addressResponse.rowCount === 0) {
-      const newAddress = await db.query(
-        'INSERT INTO address (street, number, postal_code, city, country, lat, lng) VALUES ($1, $2, $3, $4, $5, $6, $7) RETURNING address_id',
-        [street, number, postal_code, city, country, lat, lng]
+    const { base_address } = user
+    let addressId: any = base_address
+
```
MikeVedsted, line 85: type should be inferred here, so no need to overwrite it with the any type. If anything, it should be `string | Address | undefined` as type

src/services/user.ts
```diff
@@ -67,55 +67,50 @@ const updateUser = async (userId: string, update: Partial<User>) => {
...
+    if (update.address) {
+      const address: Address = update.address
+      const { street, postal_code, city, country, lat, lng } = address
+      let { number } = address
+      !number && (number = 0)
```
MikeVedsted, line 90: It should now be possible to have number be part of the destructuring on line 88 and remove line 89+90. We no longer want to save a street number as 0 if none is submitted.

client/src/redux/reducers/user.ts
```diff
@@ -0,0 +1,32 @@
+import { UserState, UserActions, ADD_USER, REMOVE_USER } from '../../types'
+
+export default function user(
+  state: UserState = {
+    user: []
+  },
```
MikeVedsted, line 6: Since only one user should be available, it would make more sense to have the UserState as an object instead of an array.

client/src/redux/reducers/user.ts
```diff
@@ -0,0 +1,32 @@
...
+    // FIX This is not implemented anywhere. Delete me on refactoring or keep me for future
+    case REMOVE_USER: {
+      const { user } = action.payload
+      const index = state.user.findIndex((p) => p.email === user.email)
+      if (index >= 0) {
+        state.user.splice(index, 1)
+        return { ...state, user: [...state.user] }
+      }
+      return state
+    }
```
MikeVedsted, line 27: If changing from array to object, these also need to be changed.

client/src/pages/ProfileEdit/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+  const { first_name, user_id } = cookies.user
+  const [user, loading] = useUser(userId)
+
+  useEffect(() => {
+    userId !== user_id && history.push(`/user/${userId}`)
+  }, [history, userId, user_id])
+
+  return (
+    !loading && (
+      <div className='page'>
```
MikeVedsted, line 23: Please don't use `page` as the class name. We are using specific naming on the other pages to avoid unintentional overlap of generic keywords and stylings (yes, AccountSetup page has the same issue but it must have slipped through..)

client/src/pages/ProfileEdit/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+const ProfileEdit = () => {
+  const { userId } = useParams<ProfilePageParamProps>()
+  let history = useHistory()
+  const [cookies] = useCookies(['user'])
+  const { first_name, user_id } = cookies.user
+  const [user, loading] = useUser(userId)
+
+  useEffect(() => {
+    userId !== user_id && history.push(`/user/${userId}`)
+  }, [history, userId, user_id])
```
MikeVedsted, line 19: Could we somehow make a different type of route protection for this? I don't mean that you should do it now but maybe make an issue for it.. We could use something similar for edit events

client/src/hooks/useUser.ts
```diff
@@ -0,0 +1,45 @@
...
+      const { data } = await axios.get(`/api/v1/users/${userId}`)
+      setUser(data)
+      dispatch(addUser(data))
+      setLoading(false)
+    } catch (error) {
+      console.log(error)
+    }
+  }
+  const fetchPublicUser = async () => {
+    try {
```
MikeVedsted, line 32: space between functions

client/src/hooks/useUser.ts
```diff
@@ -0,0 +1,45 @@
...
+      setLoading(true)
+      const { data } = await axios.get(`/api/v1/users/${userId}/public`)
+      setUser(data)
+      setLoading(false)
+    } catch (error) {
+      console.log(error)
+    }
+  }
```
MikeVedsted, line 40: I think you can also just use one function, and set the url in the ternary? 
That way, the whole function would be more DRY and would look like this:

```
const useUser = (userId: string) => {
  const dispatch = useDispatch()
  const [cookies] = useCookies(['user'])
  const { user_id } = cookies.user
  const [user, setUser] = useState<User>()
  const [loading, setLoading] = useState(true)

  useEffect(() => {
    fetchUser()
  }, [])

  const fetchUser = async () => {
    try {
      setLoading(true)
      const url =
        userId === user_id
          ? `/api/v1/users/${userId}`
          : `/api/v1/users/${userId}/public`
      const { data } = await axios.get(url)
      setUser(data)
      dispatch(addUser(data))
      setLoading(false)
    } catch (error) {
      console.log(error)
    }
  }

  return [user, loading]
}

export default useUser
```

client/src/components/AccountForm/index.tsx
```diff
@@ -8,42 +9,53 @@ import FormDropdownField from '../FormDropdownField'
...
+    first_name: user[0].first_name,
+    last_name: user[0].last_name,
+    date_of_birth: user[0].date_of_birth,
+    gender: user[0].gender,
+    profile_text: user[0].profile_text,
+    profile_image: user[0].profile_image
   })
+  const userAddress = `${user[0].street} ${user[0].number}, ${user[0].postal_code} ${user[0].city}`
+
+  useMemo(() => {
```
MikeVedsted, line 29: is useMemo the best option here? I think it should only be used for expensive functions, which setting the address isn't

MikeVedsted (commented):
I still think it is currently a weird mix of redux and no redux if we look at the app as a whole but that will probably get better when redux is more widely implemented. 

As we chatted about, your changes curently crash the app when trying to access account-setup, as there won't be a user in the state. Once that is fixed (and the current comments), I can approve your PR.

src/services/user.ts
Rostislav013: I didn't want to use `string | Address | undefined` as type, cuz it looks strange. `Any` type could be easier  spotted while refactoring types. But I'll use it then

client/src/components/GoogleUserLogin/index.tsx
```diff
@@ -11,6 +11,7 @@ const GoogleUserLogin = () => {
 
   const checkIfNew = (date: string, userId: string) => {
     const diff = new Date().getTime() - new Date(date).getTime()
+    console.log(diff)
```
MikeVedsted, line 14: Remove the console.log

client/src/components/NavDropdown/index.tsx
```diff
@@ -17,13 +19,15 @@ const NavDropdown = ({
...
+  const dispatch = useDispatch()
   const node = useRef() as React.MutableRefObject<HTMLInputElement>
   const history = useHistory()
   const [, , removeCookie] = useCookies(['user'])
 
   const logout = () => {
     removeCookie('user')
     setDropdownHidden(true)
+    dispatch(removeUser())
     history.push('/')
```
MikeVedsted, line 31: with the clickOutside implemented, you can remove line 29

client/src/pages/ProfileEdit/index.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  return (
+    !loading && (
+      <div className='editPage'>
+        <h2 className='editPage__title'>{heading}</h2>
+        <p className='editPage__text'>{paragraph}</p>
+        {user && userId === user_id && <AccountForm userId={user_id} />}
+      </div>
```
MikeVedsted, line 35: Please use kebab-case for class names

client/src/pages/ProfileEdit/ProfileEdit.scss
```diff
@@ -0,0 +1,17 @@
+@import '../../SCSS/config';
+
+.editPage {
+  padding-top: 100px;
+  display: flex;
+  flex-direction: column;
+
```
MikeVedsted, line 7: please add padding to the bottom as well, I can only see half of the submit button :)

MikeVedsted (approved):
I have some issues with the log out function in this PR. It might just be because you haven't pulled the latest changes to ProtectedRoute and refresh cookie.

client/src/pages/ProfileEdit/index.tsx
MikeVedsted: I did :)

PR #210, opened 2021-01-06
Rostislav013 refactor/landing page

MikeVedsted:
@Rostislav013 I think this is a better solution to what you were trying to achieve in #204


Chiranjibichapagain (approved):
:+1:

Rostislav013:
Yes, it is what I wanted in the end

MikeVedsted:
> Yes, it is what I wanted in the end. But then do we really wanna show how many users it has?
> In the beginning will be like 5 or 10 for ex ehh

I think after this mornings meeting, we can go ahead and implement this

PR #214, opened 2021-01-07
[FEAT] Handle 404s

MikeVedsted:
Redirect invalid routes to /get-started, which already has a redirect

### DIFF
--- a/client/src/Routes.tsx
+++ b/client/src/Routes.tsx
@@ -11,14 +11,18 @@ import MyEventsPage from './pages/MyEventsPage'
 const Routes = () => (
   <Switch>
     <ProtectedRoute exact path='/' component={HomePage} />
-    <Route exact path='/user/:userId' component={ProfilePage} />
-    <Route exact path='/get-started' component={LandingPage} />
+    <ProtectedRoute exact path='/user/:userId' component={ProfilePage} />
     <ProtectedRoute
       path='/user/:userId/:contentName'
       component={MyEventsPage}
     />
-    <Route exact path='/user/:userId/account-setup' component={AccountSetup} />
-    <Route path='/user/:userId/:contentName' component={MyEventsPage} />
+    <ProtectedRoute
+      exact
+      path='/user/:userId/account-setup'
+      component={AccountSetup}
+    />
+    <Route exact path='/get-started' component={LandingPage} />
+    <Route path='/' component={LandingPage} />
   </Switch>
 )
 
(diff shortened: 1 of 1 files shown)


Chiranjibichapagain (approved):
:+1:

client/src/Routes.tsx
```diff
@@ -11,14 +11,18 @@ import MyEventsPage from './pages/MyEventsPage'
...
     />
-    <Route exact path='/user/:userId/account-setup' component={AccountSetup} />
-    <Route path='/user/:userId/:contentName' component={MyEventsPage} />
+    <ProtectedRoute
+      exact
+      path='/user/:userId/account-setup'
+      component={AccountSetup}
+    />
+    <Route exact path='/get-started' component={LandingPage} />
+    <Route path='/' component={LandingPage} />
```
Rostislav013, line 25: Lines 13 and 25 has same path and different components

client/src/Routes.tsx
```diff
@@ -11,14 +11,18 @@ import MyEventsPage from './pages/MyEventsPage'
...
+    <ProtectedRoute exact path='/user/:userId' component={ProfilePage} />
     <ProtectedRoute
       path='/user/:userId/:contentName'
       component={MyEventsPage}
     />
-    <Route exact path='/user/:userId/account-setup' component={AccountSetup} />
-    <Route path='/user/:userId/:contentName' component={MyEventsPage} />
+    <ProtectedRoute
+      exact
+      path='/user/:userId/account-setup'
```
Rostislav013, line 21: Have u tried to open that account setup page? its path might conflict with `MyEventsPage`.
Account setup: `'/user/:userId/account-setup'`
EventsPage;      `'/user/:userId/:contentName'`
:contentName might be equal to account-setup

MikeVedsted: yes, the order of these two need to be reversed.

client/src/Routes.tsx
MikeVedsted: Although the string is the same, these two lines are not the same. And this is exactly the reason for this request.

First, line 13 is a protected route, whereas 25 is a public route.
Second, line 13 is an exact path, whereas 25 is not.
Because 25 is not exact, it will take any requets that doesn't match any of the above and send them to the landing page. The landing page has a auth check built in (built by you) which redirects users who are authenticated to the homepage instead.

Thus, any path the doesn't exist, will end on either the landing or home page, depending on the auth status.

Rostislav013 (approved):

PR #216, opened 2021-01-07
[FIX] Align links and requests

MikeVedsted:
Fix broken paths and align requests

### DIFF
--- a/client/src/components/AccountForm/index.tsx
+++ b/client/src/components/AccountForm/index.tsx
@@ -30,7 +30,7 @@ const AccountForm = ({ userId }: AccountFormProps) => {
       const update = { ...fields, address }
       const res = await axios({
         method: 'PUT',
-        url: `http://localhost:5000/api/v1/users/${userId}`,
+        url: `/api/v1/users/${userId}`,
         data: update
       })
       alert(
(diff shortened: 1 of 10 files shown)


Chiranjibichapagain (approved):
:+1:

client/src/components/AccountForm/index.tsx
```diff
@@ -30,7 +30,7 @@ const AccountForm = ({ userId }: AccountFormProps) => {
       const update = { ...fields, address }
       const res = await axios({
         method: 'PUT',
-        url: `http://localhost:5000/api/v1/users/${userId}`,
+        url: `/api/v1/users/${userId}`,
         data: update
       })
```
Rostislav013, line 35: `axios.put('`/api/v1/users/${userId}`, update)`

Rostislav013 (approved):

client/src/components/AccountForm/index.tsx
MikeVedsted: We are doing the way I've written for everything except get requests, so this is aligned. 
In my opinion, it is also more readable like this, especially if you are sending more than one variable as the data.

PR #217, opened 2021-01-07
[Feat] Add event manage options

Chiranjibichapagain:
desktop
mob

### DIFF
--- a/client/src/components/Event/Event.scss
+++ b/client/src/components/Event/Event.scss
@@ -5,6 +5,7 @@
   box-sizing: border-box;
   display: flex;
   flex-direction: column;
+  position: relative;
 
   &__data-box {
     display: flex;
@@ -16,6 +17,12 @@
     object-fit: cover;
     border-radius: 6px;
     margin-right: 25px;
+  }
+
+  &__manage {
+    color: $textDark;
+    cursor: pointer;
+    margin-right: 10px;
   }
 
   &__info-box {
@@ -96,7 +103,12 @@
 
 .title {
   display: flex;
+  justify-content: space-between;
   align-items: center;
+
+  &__wrapper {
+    display: flex;
+  }
 
   &__text {
     margin: 5px 0px 10px 0px;
--- a/client/src/components/Event/index.tsx
+++ b/client/src/components/Event/index.tsx
@@ -29,7 +29,6 @@ const Event = ({
   const [showManageOptions, setShowManageOptions] = useState(false)
   const [users] = useUserDisplay()
   const [cookies, setCookies] = useCookies(['user'])
-
   const { user_id } = cookies.user || ''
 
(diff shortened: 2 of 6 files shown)


client/src/components/Event/Event.scss
```diff
@@ -5,6 +5,7 @@
   box-sizing: border-box;
   display: flex;
   flex-direction: column;
+  position: relative;
```
MikeVedsted, line 8: Why are you adding `position: relative;` ? It doesn't look like you are actually using it

client/src/components/Event/index.tsx
```diff
@@ -24,7 +26,11 @@ const Event = ({
   handleAddRequest
 }: EventType) => {
   const [details, setDetails] = useState(false)
+  const [showManageOptions, setShowManageOptions] = useState(false)
   const [users] = useUserDisplay()
+  const [cookies, setCookies] = useCookies(['user'])
+
+  const { user_id } = cookies.user || ''
```
MikeVedsted, line 33: remove empty line 32

client/src/components/Event/index.tsx
```diff
@@ -33,13 +39,42 @@ const Event = ({
...
+  const endEvent = () => {
+    setShowManageOptions(false)
+  }
+
+  const editEvent = () => {
+    setShowManageOptions(false)
+  }
+
+  console.log('user_id', user_id, 'creator_id', eventCreator?.user_id)
+
```
MikeVedsted, line 55: remove unneeded console.log

client/src/components/Event/index.tsx
```diff
@@ -33,13 +39,42 @@ const Event = ({
...
+    setShowManageOptions(false)
+  }
+
+  const endEvent = () => {
+    setShowManageOptions(false)
+  }
+
+  const editEvent = () => {
+    setShowManageOptions(false)
+  }
```
MikeVedsted, line 52: Are you going to create these functions or create an issue for them? 😉

MikeVedsted (approved):

client/src/components/Event/Event.scss
Chiranjibichapagain: Yes, it is being used.  I needed to display the dropdown absolute but relative to the parent div (event).

client/src/components/Event/index.tsx
Chiranjibichapagain: Now they just close the dropdown, but when other components are ready like edit page participants modal etc. then we can add them.

MikeVedsted: If you are now making them now, please create an issue for them

Rostislav013 (approved):

Rostislav013:
does options close itself if I click outside of it?

PR #221, opened 2021-01-07
[Feat] Add event join request component

Chiranjibichapagain:
req

### DIFF
--- a/client/src/components/EventJoinRequest/index.tsx
+++ b/client/src/components/EventJoinRequest/index.tsx
@@ -5,12 +5,12 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
 import { EventJoinRequestProps } from '../../types'
 import './EventJoinRequest.scss'
 
-const EventJoinRequest = ({
-  user,
-  handleReject,
-  handleApprove
-}: EventJoinRequestProps) => {
+const EventJoinRequest = ({ user }: EventJoinRequestProps) => {
   const { profile_image, user_id, first_name, last_name } = user
+
+  const handleReject = () => {}
+
+  const handleApprove = () => {}
 
   return (
     <div className='participant'>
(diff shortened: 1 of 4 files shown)


client/src/components/EventJoinRequest/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+  return (
+    <div className='participant'>
+      <img
+        className='participant__image'
+        src={profile_image}
+        alt={first_name}
+      />
+      <Link className='participant__link' to={`/${user_id}`}>
+        <p className='participant__name'>{`${first_name} ${last_name}`}</p>
+      </Link>
```
MikeVedsted, line 24: the path should be `/user/${user_id}`

client/src/components/EventJoinRequest/index.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome'
+
+import { EventJoinRequestProps } from '../../types'
+import './EventJoinRequest.scss'
+
+const EventJoinRequest = ({
+  user,
+  handleReject,
+  handleApprove
+}: EventJoinRequestProps) => {
```
MikeVedsted, line 12: I think it will be better to create this functions in the component itself, rather than pass it down. The parent won't really need to know the outcome, right?

Chiranjibichapagain: umm yes! I think that's better.

Chiranjibichapagain: Then i will have another problem on how to make these events (accept and reject) happen? :D Guess, we make a new issue?

MikeVedsted: I think I just replied to this in #217 :wink

Rostislav013 (approved):

MikeVedsted (approved):

PR #222, opened 2021-01-07
[feat] Add EventList component

MikeVedsted:
Reusable listing component for events, including optional title.  Styled to match profile page but styles might need to be re-aligned across site

### DIFF
--- a/client/src/components/EventList/index.tsx
+++ b/client/src/components/EventList/index.tsx
@@ -1,7 +1,7 @@
 import React from 'react'
 
-import { EventListProps } from '../../types'
 import Event from '../../components/Event'
+import { EventListProps } from '../../types'
 import './EventList.scss'
 
 const EventList = ({ events, title }: EventListProps) => {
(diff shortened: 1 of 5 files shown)


client/src/components/EventList/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+import Event from '../../components/Event'
+import { EventListProps } from '../../types'
+import './EventList.scss'
+
+const EventList = ({ events, title }: EventListProps) => {
+  return (
+    <div className='event-list'>
+      {title && <h2 className='event-list__title'>{title}</h2>}
+      {events.map((event: any) => (
+        <Event
```
Rostislav013, line 12: Does it need here a condition for events as `events &&`? How does it act if there is no event?

Rostislav013 (approved):

Chiranjibichapagain (approved):
:+1:

client/src/components/EventList/index.tsx
MikeVedsted: If there are no events, it will just be an empty array, and it won't render anything

PR #240, opened 2021-01-11
[feat] Add public profile route

MikeVedsted:
Add route for public profile and protect route with all information


Chiranjibichapagain (approved):
:+1:

Rostislav013 (approved):

PR #243, opened 2021-01-12
[feat] refresh token

### LINKED ISSUE #220: [FEAT] Refresh expired token
Refresh expired token and black-list the expired one.

I would like to stay logged-in until I choose to log-out. 

**🚀 I'm ready for take off**  

nannadao:
Implement refresh token in private route.
To test out, remember to add `JWT_REFRESH_SECRET` variable to your .env
Currently I set the access token to expire in 5 mins so you have shorter waiting time. Later on I can extend it to 1hr.

Update 1: It is possible to refresh page and get the refreshed token now. 

Update 2: A new access token is now generated from isAuthenticated middleware. Also, if refreshToken expired, I removed the user from cookies (same with logout function). Please inform me anything you'd like to improve!

Close #220

### DIFF
--- a/client/src/hooks/useHostedEvents.tsx
+++ b/client/src/hooks/useHostedEvents.tsx
@@ -7,8 +7,8 @@ const useHostedEvents = (userId: UserId) => {
   const [hostedEvents, setHostedEvents] = useState([])
 
   useEffect(() => {
-    fetchHostedEvents()
-  }, [])
+    userId && fetchHostedEvents()
+  }, [userId])
 
   const fetchHostedEvents = async () => {
     try {
--- a/client/src/pages/ProfilePage/index.tsx
+++ b/client/src/pages/ProfilePage/index.tsx
@@ -79,7 +79,7 @@ const ProfilePage = () => {
             {`${userInfo.first_name} ${userInfo.last_name}`}
           </p>
         </div>
-        {cookies.user.user_id === userId && (
+        {currentUser === userId && (
           <div className='profile__details'>
             <h4 className='profile__details--header'> Email: </h4>
             <p className='profile__details--text'> {userInfo.email} </p>
@@ -95,7 +95,7 @@ const ProfilePage = () => {
           <h4 className='profile__details--header'> Intro: </h4>
           <p className='profile__details--text'> {userInfo.profile_text} </p>
         </div>
-        {cookies.user.user_id === userId && (
+        {currentUser === userId && (
           <div className='profile__details'>
             <h4 className='profile__details--header'> Address: </h4>
             {userInfo.base_address ? (
--- a/client/src/pages/ProfilePage/index.tsx
+++ b/client/src/pages/ProfilePage/index.tsx
@@ -14,6 +14,7 @@ const ProfilePage = () => {
   const [cookies] = useCookies(['user'])
   const { userId } = useParams<ProfilePageParamProps>()
   const [userInfo, setUserInfo] = useState({
+    user_id: '',
     profile_image: '',
     first_name: '',
     last_name: '',
@@ -28,7 +29,7 @@ const ProfilePage = () => {
     city: '',
     country: ''
   })
-  const [hostedEvents] = useHostedEvents(userId)
+  const [hostedEvents] = useHostedEvents(userInfo.user_id)
 
   useEffect(() => {
     async function getUserInfo() {
--- a/client/src/pages/ProfilePage/index.tsx
+++ b/client/src/pages/ProfilePage/index.tsx
@@ -48,7 +48,6 @@ const ProfilePage = () => {
     getUserInfo()
   }, [])
 
-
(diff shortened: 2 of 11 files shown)


src/helpers/generateToken.ts
```diff
@@ -1,16 +1,25 @@
 import jwt from 'jsonwebtoken'
 
-import { JWT_SECRET } from '../util/secrets'
-
-const generateToken = (id: string) => {
+import { JWT_SECRET, JWT_REFRESH_SECRET } from '../util/secrets'
+//Generate access Token
```
Rostislav013, line 4: Shall it be a space between there?

Rostislav013 (approved):

MikeVedsted (approved):
I think you accomplish what you set out to do with the scope of the ticket.
As improvements, you could maybe move the getUserInfo from ProtectedRoute to a hook since you are reusing the same logic in ProfilePage. 
Additionally, I  don't think we need to have two additional routes to handle this logic but rather that it could be implemented in the isAuthenticated middleware. 

I am having one issue with your request. When accessing a profile page the first time, all account values are undefined. I have to refresh first to get the actual values, regardless of my token status. You might want to check your async functions.

Your PR highlights a new issue. When the refresh token is expired, the user can currently still browse the site (just without the protected data). The user should however be taken to log in or get started, once no valid token exists.

client/src/pages/ProfilePage/index.tsx
```diff
@@ -75,38 +76,32 @@ const ProfilePage = () => {
...
           <h4 className='profile__details--header'> Full Name: </h4>
           <p className='profile__details--text'>
-            {' '}
-            {`${userInfo.first_name} ${userInfo.last_name}`}{' '}
+            {`${userInfo.first_name} ${userInfo.last_name}`}
           </p>
         </div>
-
         {cookies.user.user_id === userId && (
           <div className='profile__details'>
```
Rostislav013, line 83: Instead `cookies.user.user_id` can use `currentUser` cuz it is declared alr.

client/src/pages/ProfilePage/index.tsx
```diff
@@ -75,38 +76,32 @@ const ProfilePage = () => {
...
           </p>
         </div>
-
         <div className='profile__details'>
           <h4 className='profile__details--header'> Intro: </h4>
           <p className='profile__details--text'> {userInfo.profile_text} </p>
         </div>
-
         {cookies.user.user_id === userId && (
           <div className='profile__details'>
```
Rostislav013, line 99: currentUser

Rostislav013 (approved):

src/helpers/generateToken.ts
MikeVedsted: yes

src/helpers/generateToken.ts
```diff
@@ -1,16 +1,25 @@
...
+export const generateAccessToken = (id: string) => {
   return jwt.sign(
     {
       iss: 'JoinMe',
       sub: id
     },
     JWT_SECRET,
-    { expiresIn: '1h' }
+    { expiresIn: 60 * 60 }
```
MikeVedsted, line 12: please lower to 15 minutes

src/controllers/user.ts
```diff
@@ -13,6 +13,18 @@ export const googleLogin = async (req: Request, res: Response, next: NextFunctio
...
+    if (!req.user) {
+      throw Error
+    }
+    const { user_id } = req.user
+    return res.json(await UserService.findUserById(user_id))
+  } catch (error) {
+    next(new BadRequestError('Unexpected error', error))
+  }
+}
+
```
MikeVedsted, line 27: You don't actually need to do anything else than trigger the middleware, right? So you don't actually need this function.

MikeVedsted (approved):

src/controllers/user.ts
nannadao: This function is also to return user's info and update cookies

client/src/hooks/useHostedEvents.tsx
```diff
@@ -7,8 +7,8 @@ const useHostedEvents = (userId: UserId) => {
   const [hostedEvents, setHostedEvents] = useState([])
 
   useEffect(() => {
-    fetchHostedEvents()
-  }, [])
+    userId && fetchHostedEvents()
+  }, [userId])
```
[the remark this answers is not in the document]
nannadao: it will unfortunately trigger a loop

src/controllers/user.ts
MikeVedsted: Why not use the existing function then? You have just copied findUserById and given it a different name

nannadao: the difference is the findUserById take userId from params, here I take userId from req.user. But after having the userId, the service is the same, hence they only have different controller

PR #244, opened 2021-01-12
Feat add edit event component

Chiranjibichapagain:

### DIFF
--- a/client/src/Routes.tsx
+++ b/client/src/Routes.tsx
@@ -15,7 +15,7 @@ const Routes = () => (
     <ProtectedRoute exact path='/user/:userId' component={ProfilePage} />
     <ProtectedRoute
       exact
-      path='/eventEdit/:eventId'
+      path='/event/:eventId/edit'
       component={EventEditPage}
     />
     <ProtectedRoute
--- a/client/src/Routes.tsx
+++ b/client/src/Routes.tsx
@@ -7,11 +7,17 @@ import ProfilePage from './pages/ProfilePage'
 import HomePage from './pages/Homepage'
 import AccountSetup from './pages/AccountSetup'
 import MyEventsPage from './pages/MyEventsPage'
+import EventEditPage from './pages/EventEditPage'
 
 const Routes = () => (
   <Switch>
     <ProtectedRoute exact path='/' component={HomePage} />
     <ProtectedRoute exact path='/user/:userId' component={ProfilePage} />
+    <ProtectedRoute
+      exact
+      path='/eventEdit/:eventId'
+      component={EventEditPage}
+    />
     <ProtectedRoute
       exact
       path='/user/:userId/account-setup'
--- a/client/src/components/EventUpdateForm/index.tsx
+++ b/client/src/components/EventUpdateForm/index.tsx
@@ -1,0 +1,137 @@
+import React, { useState } from 'react'
+import axios from 'axios'
+
+import Button from '../Button'
+import InputField from '../FormInputField'
+import DropdownField from '../FormDropdownField'
+import GoogleAutoComplete from '../GoogleAutoComplete'
+import { useFormFields } from '../../hooks/useFormFields'
+import { eventCategories } from '../../util/constants/eventCategories'
+import './EventUpdateForm.scss'
+
+const EventUpdateForm = ({ data, user, eventId }: any) => {
+  const [address, setAddress] = useState({})
+  const [error, setError] = useState('')
+
+  const convertDate = (date: any) => {
+    const convertedDate = date.substr(0, 10)
+    return convertedDate
+  }
+  const date = data ? convertDate(data.date) : ''
+  const expires_at = data ? convertDate(data.expires_at) : ''
+
+  const [fields, handleFields] = useFormFields({
+    title: data ? data.title : '',
+    category: data ? data.category : '',
+    date: date,
+    time: data ? data.time : '',
+    description: data ? data.description : '',
+    maxParticipants: data ? data.max_participants : '',
+    created_by: user,
+    image: data ? data.image : '',
+    expires_at: expires_at
+  })
+
+  const handleSubmit = async (event: any) => {
+    event.preventDefault()
+    try {
+      const data = { ...fields, address }
+      await axios({
+        method: 'PUT',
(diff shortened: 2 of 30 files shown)


client/src/components/EventUpdateForm/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const convertDate = (date: any) => {
+    const convertedDate = date.substr(0, 10)
+    return convertedDate
+  }
+  const date = data ? convertDate(data.date) : ''
+  const expires_at = data ? convertDate(data.expires_at) : ''
+
+  const [fields, handleFields] = useFormFields({
+    title: data ? data.title : '',
+    category: data ? data.category : '',
```
Rostislav013, line 25: is `data ? data.title` same as `data?.title` ? Try to use it shorter way

client/src/components/EventUpdateForm/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+    date: date,
+    time: data ? data.time : '',
+    description: data ? data.description : '',
+    maxParticipants: data ? data.max_participants : '',
+    created_by: user,
+    image: data ? data.image : '',
```
Rostislav013, line 31: Try `data?.keysOfData`

client/src/components/EventUpdateForm/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+    expires_at: expires_at
+  })
+
+  const handleSubmit = async (event: any) => {
+    event.preventDefault()
+    try {
+      const data = { ...fields, address }
+      await axios({
```
Rostislav013, line 39: Did u have a problem with updating of event cuz of address?

Rostislav013 (approved):

client/src/components/EventUpdateForm/index.tsx
Chiranjibichapagain: I also need to accommodate else ' ' also there. So how's it possible?

client/src/components/EventUpdateForm/index.tsx
Chiranjibichapagain: yep. without passing address it won't update now.

client/src/components/EventUpdateForm/index.tsx
MikeVedsted: This is a ternary not an optional.
Can also be achieved with e.g. `title: data.title || ''`

client/src/Routes.tsx
```diff
@@ -7,11 +7,17 @@ import ProfilePage from './pages/ProfilePage'
...
 
 const Routes = () => (
   <Switch>
     <ProtectedRoute exact path='/' component={HomePage} />
     <ProtectedRoute exact path='/user/:userId' component={ProfilePage} />
+    <ProtectedRoute
+      exact
+      path='/eventEdit/:eventId'
+      component={EventEditPage}
+    />
```
MikeVedsted, line 20: please change the path to `/event/:eventId/edit` to align with the other paths. 
As a general improvement suggestion, would it maybe be better to have it as part of the MyEvents page, like the create event?

client/src/components/EventUpdateForm/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const expires_at = data ? convertDate(data.expires_at) : ''
+
+  const [fields, handleFields] = useFormFields({
+    title: data ? data.title : '',
+    category: data ? data.category : '',
+    date: date,
+    time: data ? data.time : '',
+    description: data ? data.description : '',
+    maxParticipants: data ? data.max_participants : '',
+    created_by: user,
```
MikeVedsted, line 30: created_by is not needed. This should not be an option to change.

client/src/components/EventUpdateForm/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const date = data ? convertDate(data.date) : ''
+  const expires_at = data ? convertDate(data.expires_at) : ''
+
+  const [fields, handleFields] = useFormFields({
+    title: data ? data.title : '',
+    category: data ? data.category : '',
+    date: date,
+    time: data ? data.time : '',
+    description: data ? data.description : '',
+    maxParticipants: data ? data.max_participants : '',
```
MikeVedsted, line 29: please rename maxParticipant to max_participants. It is currently not updating due to the wrong name

client/src/components/EventUpdateForm/index.tsx
MikeVedsted: My suggestion should fix this for both of you 😃

client/src/components/EventUpdateForm/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+    event.preventDefault()
+    try {
+      const data = { ...fields, address }
+      await axios({
+        method: 'PUT',
+        url: `/api/v1/events/${eventId}`,
+        data: data
+      })
+      setError('')
+      alert(`Success! Event: ${data.event_id} was successfully updated`)
```
MikeVedsted, line 45: could we add a history.push here?

if you do 

```
setError('')
history.push('/')
alert(`Success! Event: ${data.event_id} was successfully updated`)
```

the user will be redirected after confirming.

client/src/pages/EventEditPage/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+        setLoading(true)
+        const { data } = await axios.get(`/api/v1/events/${eventId}`)
+        setData(data)
+        setLoading(false)
+      } catch (error) {
+        console.log(error)
+      }
+    }
+    getEventInfo()
+  }, [eventId])
```
MikeVedsted, line 29: move to a hook?

MikeVedsted (approved):

MikeVedsted:
Ideally I would have liked to just use the same form as for creating events but if that is out of scope please create an issue to refactor this later

Chiranjibichapagain:
> Ideally I would have liked to just use the same form as for creating events but if that is out of scope please create an issue to refactor this later

I thought about it and as we need to use a different title and buttons text and functions, I could not think of an idea for executing that. But if that is better we can do that

PR #245, opened 2021-01-12
[refactor] Separate out queries

MikeVedsted:


src/services/request.ts
```diff
@@ -0,0 +1,17 @@
...
+    const participant = createParticipant.rows[0]
+    return { message: 'Successfully accepted', participant }
+  } catch (error) {
+    return error
+  }
+}
+
+export default {
+  acceptRequest
+}
```
Rostislav013, line 17: Does it needed to be returned as object?

MikeVedsted: What do you mean? This is how we do it for all services. There will be more functions to export here later

Rostislav013: I thought we can return like `export default acceptRequest`

Rostislav013 (approved):

src/services/request.ts
MikeVedsted: Sure, as long as there is one but there are already more in the other PRs, so no need to optimize prematurely here 😉

Chiranjibichapagain (approved):
:+1:

PR #247, opened 2021-01-13
[fix] Addition to edit Event

MikeVedsted:
@Chiranjibichapagain note that I have changed the base branch on your PR #244 Please merge yours before this one.

With these changes, I am 

- [x] removing the empty object as the inital state of the address so you won't send an empty object as address, if no new address is provided.

- [x] I corrected the error of address_id not existing on TABLE event so existing addresses can also be used. 

- [x] I refactored the address part of updateEvent slightly, so it only handles address if one is submitted. This should also solve @Rostislav013 's problem with the account edit.

### DIFF
--- a/client/src/components/Event/index.tsx
+++ b/client/src/components/Event/index.tsx
@@ -1,8 +1,7 @@
 import React, { useState } from 'react'
-import { useHistory } from 'react-router-dom'
 import axios from 'axios'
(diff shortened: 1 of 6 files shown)


Chiranjibichapagain (approved):
Tested, it works :)

Rostislav013 (approved):

PR #248, opened 2021-01-13
[style] Refactor button styling

MikeVedsted:
Add baseline border to avoid size changing when transitioning
Dry up code a bit

We won't be able to add the margin dynamically by calculating it with the current structure, so the easiest way to spread out buttons when there are several in the same container, that need to be in line, is to add (when needed)
```
& * {
... styles ...
}
```


Rostislav013 (approved):

Chiranjibichapagain (approved):
:+1:

PR #250, opened 2021-01-14
[fix] Refactor address response handling

MikeVedsted:
Fixing language bug and simplifying AutoResponse

Could those of you with device languages other than English please check if the response is now in English?


Rostislav013:
Google autocomplete is in native language, but it saves in English. I think it's fine alr

Rostislav013 (approved):

PR #246, opened 2021-01-12
[feat] Add join request handling

MikeVedsted:
# To do
## As participant
**Back end**
- [x]  Send request
- [x] Cancel request
- [x] Cancel participation
- [x] Get requests
- [x] Get participating events

**Front end**
- [x]  Send request
- [ ] Cancel request
- [ ] Cancel participation
- [x] Get requests
- [x] Get participating events

## As host
**Back end**
- [x]  Get request
- [x] Accept request
- [x] Reject request
- [x] Remove participation
- [x] Get participants for event

**Front end**
- [x]  Get request
- [x] Accept request
- [x] Reject request
- [x] Remove participation
- [x] Get participants for event

PR #269, opened 2021-01-18
Add alternate event displays

### LINKED ISSUE #262: [FEAT] Add display component for hosted events
Clearly and concisely describe the problem you are trying to solve.   
Currently, full event component is being used, including the button to request to join. 
I would like to add a component that only shows the needed elements and options.

Describe how users are affected by statements users might make or a user story.  
 Why do I have all these options? I just want an overview.

**🚀 I'm ready for take off**  

### LINKED ISSUE #263: [FEAT] Add display component for interested events
💡 I would really like to solve or include
Clearly and concisely describe the problem you are trying to solve.
Currently, full event component is being used, including the button to request to join.
I would like to add a component that only shows the needed elements and options.

👶 How would a user describe this?
Describe how users are affected by statements users might make or a user story.
Why do I have all these options? I just want an overview.

🚀 I'm ready for take off

MikeVedsted:
**Done:**
- [x] Small refactor of Event component styles to DRY up scss
- [x] Add alternative display component for hosted events
- [x]  Add alternative display component for interested events
- [x]  Add alternative display component for confirmed events

**To do:**
- [x] Stitch functionality from #246 

close #262 #263 #264


Chiranjibichapagain (approved):
:+1:

client/src/components/EventConfirmed/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+    if (id === 'comments') {
+      setHideDetails(true)
+      setHideComment(!hideComments)
+    }
+  }
+
+  const handleLeaveEvent = async () => {
+    try {
+      await axios.delete(`/api/v1/requests/${ep_id}/leave`)
+      alert('Successfully deleted')
```
Rostislav013, line 31: Just an idea for future. Shall we show alert/any to user after we get positive status for example? Or it will not show alert if for example there is no this id or any other problem which wont be caught?

client/src/pages/MyEventsPage/index.tsx
```diff
@@ -22,18 +22,24 @@ const MyEventsPage = () => {
...
+        />
       )
     if (pathname.includes('/interested'))
       return (
         <EventList
+          type='interested'
           title={'Events you have requested to join'}
           events={requestedEvents}
         />
       )
```
Rostislav013, line 38: Lines 26 and 34, 28. Which one way we should use? `type='smth'` or `type={'smth'}`

client/src/pages/MyEventsPage/index.tsx
```diff
@@ -22,18 +22,24 @@ const MyEventsPage = () => {
...
     if (pathname.includes('/confirmed'))
       return (
         <EventList
+          type='confirmed'
           events={participatingEvents}
           title={'Events you are registered for'}
```
Rostislav013, line 44: Lines 42 and 44.

Rostislav013 (approved):

client/src/components/EventConfirmed/index.tsx
MikeVedsted: We are generally handling errors horribly at the moment. Both on the front and back end. 
Most errors related to the db have status 200 in the response so it is nearly impossible to handle correctly on the front as it is now. 

But yes! We should display success and error messages more uniformly.

client/src/pages/MyEventsPage/index.tsx
MikeVedsted: Both end up the same but there is no reason to destructure a string into a string. Good catch 😄

PR #292, opened 2021-01-20
[Refactor] Refactor form input field

Chiranjibichapagain:
Add some space between label and field, remove outline while typing


Rostislav013 (approved):

client/src/components/FormInputField/InputField.scss
```diff
@@ -13,11 +13,12 @@
...
     width: 100%;
+    background-color: transparent;
     display: block;
-    border: none;
+    outline: none;
     border-bottom: 2px solid $primary;
-    margin: 0 0 20px 0;
-    padding: 5px 0;
+    margin-bottom: 25px;
+    padding-top: 10px;
```
MikeVedsted, line 21: please use mixin. This is the point of this whole exercise :)

MikeVedsted (changes requested):

client/src/components/FormInputField/InputField.scss
Chiranjibichapagain: What mixin should i use here? I am a bit confused here.

MikeVedsted: You could use the mixins `text` and the spacing constants.

PR #318, opened 2021-01-25
[refactor] Get user's data to profileEdit 

Rostislav013:
<del>We need this hook just to set user for edit and profile user pages. 
It does not need to return `user` in hook. </del>

<del>What do u think about it?
What else can be done?</del>

Now user data is taken from redux. It is setted when user auth. 
When update user, I reset user again.


MikeVedsted (commented):
I don't think we really need this hook any more. 
We will store the authenticated user in state and can get the information from there. The public data for other users, is only really needed on the profile page, so there we can just have a useEffect, that checks if the route param and current user is the same, and otherwise fetches the data.

PR #329, opened 2021-01-26
[feat] profile page

nannadao:


client/src/pages/ProfilePage/index.tsx
```diff
@@ -33,16 +42,24 @@ const ProfilePage = () => {
   const [hostedEvents] = useHostedEvents(userInfo.user_id)
```
MikeVedsted, line 42: the hosted events should always be for the requested user, so you can just use `userId` straight away

client/src/pages/ProfilePage/index.tsx
```diff
@@ -33,16 +42,24 @@ const ProfilePage = () => {
...
+          dispatch(clearErrors())
+          setUserInfo(data)
+        }
       } catch (error) {
-        console.log(error)
+        dispatch(setLoaded())
+        dispatch(setErrors(error.status, error.message))
       }
     }
     getUserInfo()
```
MikeVedsted, line 65: This needs to be revised. Since we now have the current user in the store, there is not e.g. no need for the ternary in the url. We should also only fetch if the current user is not the same as the requested user, in which case it will always be the public data that is needed.

client/src/pages/ProfilePage/index.tsx
```diff
@@ -33,16 +42,24 @@ const ProfilePage = () => {
...
             ? `/api/v1/users/${userId}`
             : `/api/v1/users/${userId}/public`
         const { data } = await axios.get(url)
-        setUserInfo(data)
+        if (data.status === 'error') {
+          throw { status: data.status, message: data.message }
+        } else {
+          dispatch(setLoaded())
+          dispatch(clearErrors())
+          setUserInfo(data)
```
MikeVedsted, line 58: since you are setting userInfo after setting loaded, it will always flash the NotFound before populating.
the correct order is:

```
setUserInfo(data)
dispatch(clearErrors())
dispatch(setLoaded())
```

client/src/pages/ProfilePage/index.tsx
```diff
@@ -33,16 +42,24 @@ const ProfilePage = () => {
...
+        }
       } catch (error) {
-        console.log(error)
+        dispatch(setLoaded())
+        dispatch(setErrors(error.status, error.message))
```
MikeVedsted, line 62: set loaded last, to avoid flashing unintended content. 
```
} catch (error) {
  dispatch(setLoaded())
  dispatch(setErrors(error.status, error.message))
 }
```

MikeVedsted (commented):
I think you can DRY up the return of `/ProfilePage/index.tsx` and the styles (which you haven't touched on this PR).
Consider refactoring the data display to a component along the lines of `EventDataBox`

client/src/pages/ProfilePage/ProfilePage.scss
```diff
@@ -90,24 +81,29 @@
...
+    max-width: 40vw;
   }
 }
 
-@media only screen and (max-width: 765px) {
+@media only screen and (max-width: 768px) {
   .profile__header {
     > * {
       width: calc(100% / 3);
     }
```
MikeVedsted, line 92: What does this do?

client/src/pages/ProfilePage/ProfilePage.scss
```diff
@@ -90,24 +81,29 @@
...
 
     &--divider-optional {
       display: block;
-      border: 1px solid $lineDim;
+      @include separator;
+    }
+
+    &--divider {
+      @include separator;
     }
```
MikeVedsted, line 107: you are setting the same style, so no need to repeat.

client/src/pages/ProfilePage/ProfilePage.scss
```diff
@@ -90,24 +81,29 @@
...
+      font-size: $sub-head;
+      margin: $smallSpacing;
       text-align: center;
     }
 
     &--divider-optional {
       display: block;
-      border: 1px solid $lineDim;
+      @include separator;
+    }
```
MikeVedsted, line 103: move include up

MikeVedsted (changes requested):
You do not need to have two separate pages. 
In your hook, you can check if the user in the store is the same as the requested one (from the params) and set the user info accordingly. 

You can then display the edit button based on the same check (maybe even make a helper function to do it or the state you already use...) as well as the other information.

I still think it would be better to have the infobox as a separate component to increase the ease of reading the page as a whole. 

As your code is currently, I am able to access public profile pages but the data does not populate.

 Make sure to pull the latest changes to scss.. You are missing a lot of new stuff on that front :)

client/src/components/NotFound/NotFound.scss
```diff
@@ -11,7 +11,7 @@
   }
 
   &__image {
-    max-width: 350px;
+    max-width: 300px;
   }
 }
```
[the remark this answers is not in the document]
nannadao: In smaller phone dimension, the image is too big it will create unnecessary scrolling

client/src/pages/ProfilePage/ProfilePage.scss
nannadao: To divide space equally between to title and 2 dividers

client/src/pages/ProfilePage/ProfilePage.scss
```diff
@@ -11,27 +10,21 @@
...
+      margin-bottom: $smallSpacing;
     }
   }
 
   &__main {
     @include flexStart;
     flex-direction: column;
-    margin-left: 40px;
+    margin-left: $largeSpacing;
     flex-basis: 70%;
```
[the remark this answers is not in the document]
nannadao: I think since the info is to the left, it makes sense to keep the events to the left as well. I can style the title to the left so it can looks more coherent with the Personal Details title?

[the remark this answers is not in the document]
MikeVedsted: Yeah, that would be fine, just looks a bit weird when some things are centered and others aren't :))

client/src/pages/ProfilePage/ProfilePage.scss
MikeVedsted: I meant the `>` selector. I've never seen it used before, so I'm just curious why, when `*` on its own should work the same

MikeVedsted (approved):

PR #340, opened 2021-01-28
Refactor event participants and requests

Chiranjibichapagain:


client/src/components/EventJoinRequest/EventJoinRequest.scss
```diff
@@ -8,7 +8,7 @@
   &__user-info {
```
MikeVedsted, line 8: Could we move the green color to the scss configuration? I think we will need a success and failure color, for when we display response messages to users.

Also, we should hide the horizontal scroll, which I think is actually caused by the hr width set in the EventParticipantsAndRequests component

client/src/components/EventParticipantsAndRequests/EventParticipantsAndRequests.scss
```diff
@@ -14,7 +12,7 @@
     width: 100%;
     height: 0.5px;
     background-color: $textDark;
```
MikeVedsted, line 14: if you set the width to be anything below 100% it won't trigger the horizontal scroll

MikeVedsted (approved):

